PR #4, opened 2018-07-12
AtomicBoolean with tests

SokolovaMaria:


atomicfu-common/src/main/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
```diff
@@ -120,6 +131,79 @@ public inline fun <T> AtomicRef<T>.updateAndGet(function: (T) -> T): T {
...
+/**
+ * Atomic reference to a [Boolean] variable with volatile reads/writes via
+ * [value] property and various atomic read-modify-write operations
+ * like [compareAndSet] and others.
+ */
+public expect class AtomicBoolean {
+    /**
+     * Reading/writing this property maps to read/write of volatile variable.
+     */
+    public var value: Int
```
qwwdfsad: It should be `Boolean`

This code should compile:
```
val flag = atomic(false)
flag.value = true
```

SokolovaMaria: Thanks 
Fixed.

elizarov:
(Merged from `atomic-boolean` branch)

PR #12, opened 2018-08-09
Atomicfu plugin

SokolovaMaria:


elizarov:
New plugin breaks access to `internal` from tests. Checkout out this simple demo project: https://github.com/elizarov/AtomicFuPluginPlayground

SokolovaMaria:
Kotlin gradle plugin adds original `mainSourceSet.output` directories as friend dependency to classPath of `compileTestKotlin` task (to make it possible to use internal methods from the main source set in test classes). So, we can just replace original output, added to `compileTestKotlin` classPath, with transformed one at the beginning of `compileTestKotlin` task.

elizarov:
I've update readmes. Please pull from `plugin` branch before making additional changes.

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -47,6 +60,13 @@ fun Project.configureTransformation() {
     }
 }
 
+fun Jar.setupJarManifest(classifier: String = "") {
+    this.classifier = classifier
+    manifest.attributes.apply {
```
elizarov: This needs to be set only in `BOTH` variant.

elizarov:
I've collected everything into a single commit and rebased onto master.

PR #19, opened 2018-08-16
Tracing atomic operations

SokolovaMaria:


atomicfu-js/src/main/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,20 @@
+package kotlinx.atomicfu
+
+public actual class Trace actual constructor (size: Int) {
+    private val mask = size - 1
+    private val a = arrayOfNulls<String>(size)
+    private val index = atomic(0)
+
+    init {
+        check(size and mask == 0) { "size must be power of 2: $size"}
```
qwwdfsad: it's usually worth to round up to the next power of two instead of errors `1 << (32 - Integer.numberOfLeadingZeros(value - 1));`

atomicfu-common/src/main/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
```diff
@@ -217,6 +221,8 @@ public expect class AtomicInt {
      */
     public var value: Int
 
+    public val trace: Trace?
```
qwwdfsad: can it be non-nullable to avoid `?.` everywhere?
Default no-op implementation should be fine

atomicfu-common/src/main/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,7 @@
+package kotlinx.atomicfu
+
+public expect class Trace(size: Int) {
+    public fun append(text: String)
```
qwwdfsad: @elizarov are we going to opt-in for the logging format by default?
E.g. K/N has something like this in the compiler:
```
val foo = ...
DEBUG {
  // check some expensive foo invariant
}

```

qwwdfsad: Also, it definitely should be `open`

settings.gradle
```diff
@@ -3,7 +3,7 @@ enableFeaturePreview('GRADLE_METADATA')
 include 'atomicfu'
 include 'atomicfu-common'
 include 'atomicfu-js'
-include 'atomicfu-native'
+//include 'atomicfu-native'
```
qwwdfsad: oops

atomicfu-common/src/main/kotlin/kotlinx/atomicfu/Trace.kt
elizarov: It is not so simple. The references to the `Trace` class get erased by the bytecode post-prossor, so you cannot easily really customize it by creating your own subclass. Transformer should have to be smart enough to erase your custom subclasses and their constructors, too. Let's finish (and cleanup) basic implementation first and open a separate issue for that. See #20

settings.gradle
elizarov: TBD. Native needs to be supported before we can merge it to master.

atomicfu-common/src/main/kotlin/kotlinx/atomicfu/Trace.kt
elizarov: See #20 for custom formats

elizarov: `fun append` shall be `@PublishedApi internal`, because bytecode processor only support code patterns that result from its usage from inside of `operator fun invoke`. Direct invocations of `append` in client code are not supported.

atomicfu-common/src/main/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,7 @@
+package kotlinx.atomicfu
+
+public expect class Trace(size: Int) {
```
elizarov: Needs KDocs on class and all its functions.

atomicfu-js/src/main/kotlin/kotlinx/atomicfu/Trace.kt
elizarov: Agree. Let's round up to the next power of two.

atomicfu-common/src/main/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
```diff
@@ -73,6 +73,8 @@ public expect class AtomicRef<T> {
      */
     public var value: T
 
+    public val trace: Trace?
```
elizarov: Atomic classes shall NOT expose `trace` field. Please, remove. The correct usage pattern for trace is this:

```
val trace = Trace() // a separate field for trace
val a = atomic(0, trace)
```

Then later in code you can do `trace { "something" }`.

Bytecode transformer shall completely remove `trace` field from the source class, all its accessors, and references.

atomicfu-common/src/main/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
elizarov: The trace field shall not be exposed at all. If you need to trace, you explicitly create your own trace instance with `val trace = Trace()` in your class.

atomicfu-common/src/main/kotlin/kotlinx/atomicfu/Trace.kt
elizarov: Note, you shall also use `@JvmName("append")` in additional to marking it as `internal`, so that there is a stable JVM name you can match in bytecode processor.

SokolovaMaria:
[Trace](https://trello.com/c/mby9VL0U/25-trace)

SokolovaMaria:
[Trace](https://trello.com/c/mby9VL0U/25-trace)

elizarov:
Superceded by #40

PR #33, opened 2018-10-01
Fix #32: Use scoped this

### LINKED ISSUE #32: Correct scope for inline functions
I'm experiencing a deadlock using `kotlin.coroutines` Mutex (0.26.0 with 1.2 and EAP with Kotlin 1.3). https://github.com/ptmt/kotlinjsplayground/blob/master/src/main/kotlin/MutexCase.kt

In particular, this line of code doesn't work as expected:
https://github.com/Kotlin/kotlinx.coroutines/blob/4cc0e294b39ae0f38cafd49187f399b0ecc2e615/common/kotlinx-coroutines-core-common/src/sync/Mutex.kt#L472

```js
(affected as MutexImpl)._state.compareAndSet(this@UnlockOp, update)
```

It gets compiled into following JavaScript:

(somewhere inside `kotlinx-coroutines-core.js`)
```js
(function(scope) {
    return (Kotlin.isType(tmp$ = affected, MutexImpl) ? tmp$ : throwCCE())._state_0 === this ? function() {
        (Kotlin.isType(tmp$ = affected, MutexImpl) ? tmp$ : throwCCE())._state_0 = update;
        return true;
    }() : false
})(this);
```
*(there is also some unnecessary double-assignment but that's not the point)*

As we can see `affected` is being compared with the wrong `this`, because we have a closure here. I believe it should be replaced with `scope`. As far as I understand it's happening here:

https://github.com/Kotlin/kotlinx.atomicfu/blob/145542d596c4767dc2bec85551b4e82c2eb7b38b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt#L208-L212

Probably it could be improved by adding additional checks for this or getting rid of closure. I'd contribute a PR by myself, but I'm afraid that writing proper tests would take much longer than fix itself.

ptmt:
This is a naive incomplete fix for #32

### DIFF
--- a/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
+++ b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
@@ -203,63 +203,63 @@ class AtomicFUTransformerJS(
         val code = when (funcName) {
             "getAndSet\$atomicfu" -> {
                 val arg = args[0].toSource()
-                "(function(scope) {var oldValue = $f; $f = $arg; return oldValue;})"
+                "(function($SCOPE) {var oldValue = $f; $f = $arg; return oldValue;})"
             }
             "compareAndSet\$atomicfu" -> {
-                val expected = args[0].toSource()
+                val expected = args[0].scopedSource()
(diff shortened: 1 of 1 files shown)


elizarov:
Incorporated into #34. Thanks a lot!

PR #35, opened 2018-10-02
Some fragilities of JS transformer fixed (+ fixed JS local testing)

### LINKED ISSUE #25: Fragile JS transformer
AtomicFU js transformer is fragile. It works for `kotlinx.coroutines` but it is not generally applicable to wider range of libraries. Here is a snippet of produced JS code:

```
(function(root, factory) {
  if (typeof define === 'function' && define.amd) 
    define(['exports', 'kotlin'], factory);
  else if (typeof exports === 'object') 
    factory(module.exports, require('kotlin'));
  else {
    if (typeof kotlin === 'undefined') {
      throw new Error("Error loading module 'kotlinx-coroutines-core'. Its dependency 'kotlin' was not found. Please, check whether 'kotlin' is loaded prior to 'kotlinx-coroutines-core'.");
    }
    if (typeof this['kotlinx-atomicfu'] === 'undefined') {
    }
    root['kotlinx-coroutines-core'] = factory(typeof this['kotlinx-coroutines-core'] === 'undefined' ? {} : this['kotlinx-coroutines-core'], kotlin);
  }
}(this, function(_, Kotlin, $module$kotlinx_atomicfu) {
  'use strict';
...
  var atomic;
  var atomic_0;
....
  $$importsForInline$$['kotlinx-atomicfu'] = $module$kotlinx_atomicfu;
```

Things that should be fixed:
1. ` if (typeof this['kotlinx-atomicfu'] === 'undefined') { ... }` should be completely removed (not just it body), while maintaining line-numbers (fill with appropriate number of blank lines or comments)
2. ` function(_, Kotlin, $module$kotlinx_atomicfu) { ... ` header should be corrected, its `$module$kotlinx_atomicfu` parameters shall be removed at the appropriate position (note, that it may receive additional modules _after_ `kotlinx_atomicfu`)
3. `var atomic;` and `var atmoic_0;` shall be removed completely (replaced with a blank like), not just their right-hand side.
3. ` $$importsForInline$$['kotlinx-atomicfu'] = $module$kotlinx_atomicfu;` line shall be replaced with a blank like (if present).

SokolovaMaria:
Fixes #25 
The [problem](https://github.com/Kotlin/kotlinx.atomicfu/pull/28#issuecomment-425624576) was that kotlinx-atomicfu module was not passed during local JS testing.

### DIFF
--- a/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
+++ b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
@@ -13,6 +13,7 @@ private const val SCOPE = "scope"
 private const val FACTORY = "factory"
 private const val REQUIRE = "require"
 private const val KOTLINX_ATOMICFU = "'kotlinx-atomicfu'"
+private const val MODULE_KOTLINX_ATOMICFU = "\$module\$kotlinx_atomicfu"
 
 class AtomicFUTransformerJS(
     inputDir: File,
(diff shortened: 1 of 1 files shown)


elizarov:
Looks good!

PR #65, opened 2019-04-29
Getting array elements by named index fixed

SokolovaMaria:


atomicfu/src/commonTest/kotlin/kotlinx/atomicfu/test/AtomicArrayTest.kt
```diff
@@ -156,4 +156,20 @@ class AtomicArrayWithMethod {
         val result = refArray[index].compareAndSet(null, data)
         if (!result) error("Double set")
     }
+}
+
+class IndexArrayElementGetter {
+    val clazz = AtomicArrayClass()
+    fun intGet(index: Int) = { clazz.longArr[index].value }
```
elizarov: Something is not right here. Probably extra curly braces.

atomicfu/src/commonTest/kotlin/kotlinx/atomicfu/test/AtomicArrayTest.kt
```diff
@@ -156,4 +156,20 @@ class AtomicArrayWithMethod {
...
+
+    @Suppress("NOTHING_TO_INLINE")
+    inline fun refGet(index1: Int): Any? = clazz.refArr[index1].value
+
+    @Test
+    fun testIndexArrayElementGetting() {
+        val value = intGet(3)
+        val ref = refGet(2)
```
elizarov: This test does not actually test that `value` and `ref` actually work and produces the correct value.

atomicfu/src/commonTest/kotlin/kotlinx/atomicfu/test/AtomicArrayTest.kt
```diff
@@ -156,4 +156,20 @@ class AtomicArrayWithMethod {
...
+        val elementIndex1 = 2
+        clazz.intArr[elementIndex1].value = 55
```
elizarov: This test does not actually test that assignment actually goes through. Transformer can replace this assignment with NOP and the test will still pass.

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
```diff
@@ -66,8 +67,10 @@ class AtomicFUTransformerJS(
     private fun AstNode.eraseGetValue(): String {
         val res = this.toSource()
         val primitiveGetValue = MANGLE_VALUE_REGEX
-        val arrayGetValue = ARRAY_GET_VALUE_REGEX
-        return res.replace(arrayGetValue) { matchResult -> "[${matchResult.groupValues[1]}]" }
+        val arrayGetElement = ARRAY_GET_ELEMENT_REGEX
+        val arrayGetElementValue = ARRAY_GET_ELEMENT_VALUE_REGEX
+        return res.replace(arrayGetElementValue) { matchResult -> "[${matchResult.groupValues[1]}]" }
+                  .replace(arrayGetElement) { matchResult -> "[${matchResult.groupValues[1]}]" }
```
elizarov: This does not look right. It is bound to fail in cases like `array[i + j].value`, e.g. when there is complex expression in place of an index. Please, add the corresponding tests, too.

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
```diff
@@ -26,8 +26,8 @@ private const val FILL = "fill"
 private const val GET_ELEMENT = "get\\\$atomicfu"
 
 private val MANGLE_VALUE_REGEX = Regex(".${Pattern.quote(MANGLED_VALUE_PROP)}")
-private val ARRAY_GET_ELEMENT_REGEX = Regex(".$GET_ELEMENT\\(([A-Za-z0-9$]+)\\)")
-private val ARRAY_GET_ELEMENT_VALUE_REGEX = Regex("$ARRAY_GET_ELEMENT_REGEX.${Pattern.quote(MANGLED_VALUE_PROP)}")
+// matches index until the first occurence of ')', parenthesised index expressions not supported
+private val ARRAY_GET_ELEMENT_REGEX = Regex(".$GET_ELEMENT\\(([^)]*)\\)")
```
SokolovaMaria: For now, only simple non-parenthesised expressions are supported as an index.
Can I make a recursive regular expression like this: 
`\((?>[^()]|(?R))*\)` that would match a single pair of parentheses with any symbols in between including balanced parentheses to support more complex expressions:
`arr[atomic1.value]`, `arr[f(i) + f(j)]`..?

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
```diff
@@ -64,11 +65,32 @@ class AtomicFUTransformerJS(
 
     // erase getting value of atomic field
     private fun AstNode.eraseGetValue(): String {
-        val res = this.toSource()
+        var res = this.toSource()
```
elizarov: You have `AstNode` here. Why do you really need to convert it to string a parse it again via regeexps?

atomicfu/src/commonTest/kotlin/kotlinx/atomicfu/test/AtomicArrayTest.kt
```diff
@@ -156,4 +156,24 @@ class AtomicArrayWithMethod {
         val result = refArray[index].compareAndSet(null, data)
         if (!result) error("Double set")
     }
+}
+
+class IndexArrayElementGetter {
```
elizarov: Style: separate test classes shall be in separate files and named as `xxxTest`

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
SokolovaMaria: In the AST I use for transformation, it's possible to transform nodes only without changing their type. "Get value" nodes like these: 
```
arr.get$atomicfu(3).kotlinx$atomicfu$value
x.kotlinx$atomicfu$value
```
are of type PropertyGet: `x` is a target field, `kotlinx$atomicfu$value` is a property field and after transformation, it should be of type Name: just `x`. So, I extract all property get calls via regeexps now.

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
elizarov: Will defer to a later time: https://github.com/Kotlin/kotlinx.atomicfu/issues/71

elizarov:
Note. I've squashed it all into a single commit and force-pushed. It still does not work. Take a look at `semaphore` branch of `kotlinx.coroutines` -- it still fails JS tests with this version of atomicfu.

Issue #67, opened 2019-05-01
broken publish to npm, from version 0.12.0


beyondeye:
Apparently all versions from version 0.12.0 and later has broken publishing to npm:
 only ```package.json``` and ```README.md``` are present in kotlinx-atomicfu subdirectory in node_modules, after installing it with 
```
npm i kotlinx-atomicfu
```

elizarov:
Fixed in `0.12.7`

PR #69, opened 2019-05-05
fix for broken npm publishing

beyondeye:
Fixes issue #67


elizarov:
Thanks! I've merged that into `develop` branch for upcoming release.

PR #74, opened 2019-08-16
Add FQN for atomicfu plugin

Prototik:
To be convenient with default Kotlin gradle plugin and Gradle's naming conventions for plugins I suggest to add full-qualified name for atomicfu plugin.
It can be used in the following ways:
```kotlin
plugins {
  id("org.jetbrains.kotlin.jvm") version "${kotlin_version}"
  id("org.jetbrains.kotlin.plugin.atomicfu") version "${atomicfu_version}"
}
```
Or via Gradle Kotlin DSL extensions:
```kotlin
plugins {
  kotlin("jvm") version "${kotlin_version}"
  kotlin("plugin.atomicfu") version "${atomicfu_version}"
}
```
I choose the prefix `org.jetbrains.kotlin.plugin` similarly to `org.jetbrains.kotlin.plugin.serialization` from `kotlinx.serialization`

### DIFF
--- a/atomicfu-gradle-plugin/src/main/resources/META-INF/gradle-plugins/org.jetbrains.kotlin.plugin.atomicfu.properties
+++ b/atomicfu-gradle-plugin/src/main/resources/META-INF/gradle-plugins/org.jetbrains.kotlin.plugin.atomicfu.properties
@@ -1,0 +1,1 @@
+implementation-class=kotlinx.atomicfu.plugin.gradle.AtomicFUGradlePlugin


elizarov:
Nice job. I think that the README should be adjusted correspondly, too.

Prototik:
I'm not sure about description in readme, as atomicfu doesn't published to gradle plugin portal.
For now I'm using following snippet in `settings.gradle.kts`:
```kotlin
val atomicfu_version: String by settings

pluginManagement {
    repositories {
        maven("https://kotlin.bintray.com/kotlinx") {
            content {
                includeGroup("org.jetbrains.kotlinx")
            }
        }
    }

    resolutionStrategy {
        eachPlugin {
            with(requested.id) {
                when {
                    id == "kotlinx-atomicfu" ->
                        useModule("org.jetbrains.kotlinx:atomicfu-gradle-plugin:$atomicfu_version")
                }
            }
        }
    }
}
```
(actually this don't cannot be changed until atomicfu published to plugin portal, except of plugin id)
Should I add that snippet to readme or just mention that instead `kotlinx-atomicfu` can be used FQN notation?

SokolovaMaria (approved):

PR #75, opened 2019-08-29
array-element-inline-func

SokolovaMaria:
Inline functions on array elements fixed

### DIFF
--- a/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
+++ b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
@@ -676,14 +676,8 @@ class AtomicFUTransformer(
             }
             // An operation other than getValue/setValue is used
             if (f.isArray && iv.name == "get") { // "operator get" that retrieves array element, further ops apply to it
-                // save stack start of atomic operation args
-                val args = iv.next
-                // remove getter of array element
-                instructions.remove(iv)
                 // fixup atomic operation on this array element
-                val nextAtomicOperation = FlowAnalyzer(args).execute() as MethodInsnNode
-                val fixedAtomicOperation = fixupInvokeVirtual(ld, true, nextAtomicOperation, f)
-                return fixedAtomicOperation!!.next
+                return fixupLoadedArrayElement(f, ld, iv)
             } else {
                 // non-trivial atomic operation
                 check(f.isArray == onArrayElement) { "Atomic operations can be performed on atomic elements only" }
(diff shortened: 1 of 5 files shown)


atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -785,33 +775,71 @@ class AtomicFUTransformer(
...
+        private fun fixupLoadedArrayElement(f: FieldInfo, ld: FieldInsnNode, getter: MethodInsnNode): AbstractInsnNode? {
+            // contains array field load (in vh case: + swap and pure type array load) and array element index
+            val arrayElementInfo = mutableListOf<AbstractInsnNode>()
+            if (vh) {
+                arrayElementInfo.add(ld.previous.previous) // getstatic VarHandle field
+                arrayElementInfo.add(ld.previous) // swap
+            }
+            var i: AbstractInsnNode = ld
+            while (i != getter) {
+                arrayElementInfo.add(i)
```
elizarov: This is very brittle. It will not work with complex expression in array access, like `arr[<complex-expression>].update { ... }`. What's worse, with code code `arr[funCall()].update { ... }`, the code for `funCall()` will get duplicated and code semantics will change.

Please, add more diagnostics here. Similarly to data-flow analyzer, please ensure that this piece of code that is being stored in `arrayElementInfo` does not contain any complex data-flow. Moreover, it should not contain any invocations. Only simple loads and simple arithmetics can be allowed here. If anything else is encountered, then please complain with an appropriate message.

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -785,33 +775,71 @@ class AtomicFUTransformer(
...
+            // start of array element operation arguments
+            val args = getter.next
+            // remove array element getter
+            instructions.remove(getter)
+            val arrayElementOperation = FlowAnalyzer(args).execute()
+            return fixupOperationOnAtomicVar(arrayElementOperation, f, ld, arrayElementInfo)
+        }
+
+        private fun fixupOperationOnAtomicVar(operation: AbstractInsnNode, f: FieldInfo, ld: FieldInsnNode, arrayElementInfo: List<AbstractInsnNode>?): AbstractInsnNode? {
+            val onArrayElement = f.getPrimitiveType(vh).sort == ARRAY
```
elizarov: It seems redundant now, as `onArrayElement` shoud be equivalent to`arrayElementInfo != null`
At least, an assertion to this fact would be helpful here:
```
check(onArrayElement == (arrayElementInfo != null))
```
However, when I add this kind of `check` it fails on `testIndexArrayElementGetting`. This is strage. It means that some loads are not being recognized as "array loads", so they work like regular functions, but will not work with `inline` function. 

It looks like the problem with primitive arrays with inline functions. Please, add corresponding tests.

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -785,33 +775,71 @@ class AtomicFUTransformer(
...
                         lv.desc = f.ownerType.descriptor
                         lv.signature = null
                         // process all loads of this variable in the corresponding local variable range
                         forVarLoads(v, lv.start, lv.end) { otherLd ->
-                            fixupVarLoad(f, ld, otherLd)
+                            if (!onArrayElement)
+                                fixupVarLoad(f, ld, otherLd)
+                            else {
+                                fixupArrayElementLoad(f, ld, otherLd, arrayElementInfo!!)
+                            }
```
elizarov: This whole `if (!onArrayElement) { fixupVarLoad(...) } else {  fixupArrayElementLoad(...) }` is copied in two places of code. It is better to extract it into some kind of `fixupLoad` method. 

Also, style note: prefer writing `if (onArrayElement) ...` (direct form) as opposed to `if (! onArrayElement) ...` (negated). There is a refactoring to "invert if".

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
SokolovaMaria: `fixupLoadedArrayElement` is called only when `get` array element instruction is visited: it gets information about array field and element index (`arrayElementInfo`) that will be used only in case of loading a stored reference to this array element. 
So, `arrayElementInfo` is not null only if operation on the array element is being processed but if this fixup is processing operation on the array field itself (only the `get(I)` element instruction can follow it ) then `arrayElementInfo` is still null. Though,  in both cases `f.getPrimitiveType(vh).sort == ARRAY`. 

I moved this check for `arrayElementInfo != null` to the branch of storing operation as it is only used in that case.

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -785,46 +775,126 @@ class AtomicFUTransformer(
...
         }
 
+        private fun fixupArrayElementLoad(f: FieldInfo, ld: FieldInsnNode, otherLd: VarInsnNode, arrayElementInfo: List<AbstractInsnNode>): AbstractInsnNode? {
+            // index instructions from array element info: drop owner class load instruction (in vh case together with preceding getting VH + swap)
+            val index = arrayElementInfo.drop(if (vh) 3 else 1)
+            // previously stored array element reference is loaded -> arrayElementInfo should be cloned and inserted at the point of this load
+            // before cloning make sure that index instructions contain just loads and simple arithmetic, without any invocations and complex data flow
+            for (indexInsn in index) {
+                checkDataFlowComplexity(indexInsn)
+            }
```
elizarov: Good. Can we just do this verification loop in `fixupLoadedArrayElement` right where this `arrayElementInfo` is created?

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -806,46 +800,126 @@ class AtomicFUTransformer(
...
                 is MethodInsnNode -> {
                     // invoked virtual method on atomic var -- fixup & done with it
-                    debug("invoke $f.${j.name}", sourceInfo.copy(i = j))
-                    return fixupInvokeVirtual(ld, false, j, f)
+                    debug("invoke $f.${operation.name}", sourceInfo.copy(i = operation))
+                    return fixupInvokeVirtual(ld, arrayElementInfo != null, operation, f)
                 }
                 is VarInsnNode -> {
+                    val onArrayElement = arrayElementInfo != null
+                    check((f.getPrimitiveType(vh).sort == ARRAY) == (arrayElementInfo != null))
```
elizarov: Style nit: `check(f.isArray == onArrayElement)` would look cleaner here.

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -806,46 +800,126 @@ class AtomicFUTransformer(
...
+                        fixupLoad(f, ld, nextVarLoad(v, next), arrayElementInfo)
                     }
                     return next
                 }
                 else -> abort("cannot happen")
             }
         }
 
+        private fun fixupLoad(f: FieldInfo, ld: FieldInsnNode, otherLd: VarInsnNode, arrayElementInfo: List<AbstractInsnNode>?): AbstractInsnNode? {
+            return if (arrayElementInfo != null) {
```
elizarov: Style nit: This better looks as expression function (can be fixed with Alt+ENTER -> Convert to expression body:
```
private fun fixupLoad(...) = 
    if (....) ....
```

PR #76, opened 2019-08-30
Unchecked casts erased

SokolovaMaria:
AtomicRef val is never null, so for JVM I can rely that nonnull check may be inserted by the compiler only and check for `dup, nonnull` pattern and then erase all instructions of null branch.

For JS unchecked casts like `(Kotlin.isType(tmp$_0 = atomicField, AtomicRef)` are erased and replaced with just `atomicField`

### DIFF
--- a/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
+++ b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
@@ -28,6 +28,8 @@ private val INT_ARRAY_TYPE = getType("[I")
 private val LONG_ARRAY_TYPE = getType("[J")
 private val BOOLEAN_ARRAY_TYPE = getType("[Z")
 private val REF_ARRAY_TYPE = getType("[Ljava/lang/Object;")
+private val REF_TYPE = getType("L$AFU_PKG/AtomicRef;")
+private val ATOMIC_ARRAY_TYPE = getType("L$AFU_PKG/AtomicArray;")
 
 private val AFU_CLASSES: Map<String, TypeInfo> = mapOf(
     "$AFU_PKG/AtomicInt" to TypeInfo(getObjectType("$JUCA_PKG/AtomicIntegerFieldUpdater"), INT_TYPE, INT_TYPE),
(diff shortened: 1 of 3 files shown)


atomicfu/src/commonTest/kotlin/kotlinx/atomicfu/test/UncheckedCastTest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @Test
+    fun testAtomicValUncheckedCast() {
+        assertEquals((s as AtomicRef<String>).value, "AAA")
+        bs.lazySet(mapOf(1 to listOf(Box(1), Box(2))))
+        assertEquals((bs as AtomicRef<Map<Int, List<Box>>>).value[1]!![0].b * 10, 10)
+    }
+
+    @Test
+    fun testTopLevelValUnchekedCast() {
```
elizarov: typo 'Uncheked' -> 'Unchecked'

atomicfu/src/commonTest/kotlin/kotlinx/atomicfu/test/UncheckedCastTest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+private val topLevelS = atomic<Any>(listOf("A", "B"))
+
+class UncheckedCastTest {
+    private val s = atomic<Any>("AAA")
+    private val bs = atomic<Any?>(null)
+
+    @Test
+    fun testAtomicValUncheckedCast() {
+        assertEquals((s as AtomicRef<String>).value, "AAA")
```
elizarov: Please, add a test on using it with `inline` function. It works perfectly now, but a test would still be extremely helpful to have for future maintenance. Add something like this:

```
    inline fun <T> AtomicRef<T>.getString(): String =
        (this as AtomicRef<String>).value
```

The test line here:

```
        assertEquals(s.getString(), "AAA")
```

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -784,6 +785,18 @@ class AtomicFUTransformer(
         }
 
         private fun fixupLoadedAtomicVar(f: FieldInfo, ld: FieldInsnNode): AbstractInsnNode? {
+            if (f.fieldType == REF_TYPE && ld.next.opcode == DUP && ld.next.next.opcode == IFNONNULL) {
```
elizarov: This check seems to be very restrictive and the code does not work with array values. Consider the following test:

```
    private val a = atomicArrayOfNulls<Any?>(10)

    @Test
    fun testArrayValueUncheckedCast() {
        a[0].value = "OK"
        assertEquals("OK", (a[0] as AtomicRef<String>).value)
    }
```

It produces Ok diagnostics ("Unsupported operation on atomic variable"). However, the following test:

```
    private val a = atomicArrayOfNulls<Any?>(10)

    @Test
    fun testArrayValueUncheckedCast() {
        a[0].value = "OK"
        assertEquals("OK", a[0].getString())
    }

    inline fun <T> AtomicRef<T>.getString() =
        (this as AtomicRef<String>).value
```

Completely crashes transformer: 

```
[main] ERROR kotlinx.atomicfu.transformer.AtomicFUTransformer - UncheckedCastTest.kt: kotlinx.atomicfu.test.UncheckedCastTest::testArrayValueUncheckedCast: Failed to transform: java.lang.ClassCastException: org.objectweb.asm.tree.VarInsnNode cannot be cast to org.objectweb.asm.tree.MethodInsnNode
java.lang.ClassCastException: org.objectweb.asm.tree.VarInsnNode cannot be cast to org.objectweb.asm.tree.MethodInsnNode
        at kotlinx.atomicfu.transformer.AtomicFUTransformer$TransformerMV.fixupInvokeVirtual(AtomicFUTransformer.kt:671)
        at kotlinx.atomicfu.transformer.AtomicFUTransformer$TransformerMV.fixupLoadedAtomicVar(AtomicFUTransformer.kt:805)
...
```

It seems a little bit of extra effort to generalize this code a bit so that casting of array elements is supported, too.

atomicfu/src/commonTest/kotlin/kotlinx/atomicfu/test/UncheckedCastTest.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlin.test.assertTrue
+
+private val topLevelS = atomic<Any>(listOf("A", "B"))
+
+class UncheckedCastTest {
+    private val s = atomic<Any>("AAA")
+    private val bs = atomic<Any?>(null)
+
+    @Test
+    fun testAtomicValUncheckedCast() {
```
elizarov: General test style: `assertEquals` first parameters is "expected value" and the second is "actual", so checks here should be written like this:
```
assertEquals( "AAA", (s as AtomicRef<String>).value,)
```
Let's write them correctly in new tests, will fix old ones later.

PR #93, opened 2019-10-08
Extend the fix from f404e552 to single-platform projects

h0tk3y:
The fix was mistakenly only applied to multiplatform projects. Extend it to
single-platform projects as well. Also, fix the compiler argument for the JS
compiler, which accepts -Xfriend-modules, not -Xfriend-paths.


qwwdfsad:
@SokolovaMaria please review and release (after merging and before releasing please check that `kotlinx-train` compiles)

PR #95, opened 2019-10-14
Native implementation of synchronized object

SokolovaMaria:


ndkoval:
In the current version, the size of the pool equals to the maximal number of required at the same time mutexes (not now but at some point). Therefore, if you don't need so many mutexes for the rest of the execution, you practically have memory leaks, because these extra mutexes are already allocated and stored in the pool. To solve this problem, I suggest NOT returning mutexes to the pool with some small probability (e.g., 5%). Thus, the number of mutexes decreases over time if you don't need all of them, and you avoid memory leaks in some sense. The cost of this solution is cheap in practice since you just have to acquire 5% extra mutexes on average while you usually do not even need them (see the implemented optimization for non-contended scenarios).

elizarov:
The overall observation here is that "reentrant" part of this lock is not tested at all. It looks like reentrancy can be added to the main stress-test, too. By changing each test thread so that it takes a lock reentrantly random number of times (from 1 to 3 is enough). It should quickly surface all bugs with reentrancy.

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/Synchronized.kt
```diff
@@ -0,0 +1,204 @@
...
+                            return
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    public fun unlock() {
+        val currentThreadId = pthread_self()
```
elizarov: This does not compile on MacOS. You need `pthread_self()!!` (it is nullable there). There will be warning during Linux compilation, though. You can avoid it by doing a custom wrapper.

atomicfu/src/nativeTest/kotlin/kotlinx/atomicfu/test/SynchronizedTest.kt
```diff
@@ -0,0 +1,38 @@
...
+                        val oldValue = count.value
+                        count.value = oldValue + 1
+                        repeat(nestedLocks) { lock.unlock() }
+                    }
+                }
+            }
+            workers.forEach {
+                it.requestTermination().result
+            }
+            assertEquals(counter.value, COUNT * 1000)
```
elizarov: `assertEquals` takes `expect` as first parameter, `actual` as second. It is reversed in this call, making error messages wrong.

atomicfu/src/nativeTest/kotlin/kotlinx/atomicfu/test/SynchronizedTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import kotlin.test.*
+
+class SynchronizedTest {
+
+    private val iterations = 1000
+    private val COUNT = 20
+
+    @Test
+    fun stressCounterTest() {
+        repeat(iterations) { i ->
```
elizarov: Parameter `i` is never used.

atomicfu/src/nativeTest/kotlin/kotlinx/atomicfu/test/SynchronizedTest.kt
```diff
@@ -0,0 +1,38 @@
...
+            val workers = Array(COUNT) { Worker.start() }
+            val counter = AtomicInt(0).freeze()
+            val so = SynchronizedObject().freeze()
+            workers.forEach { worker ->
+                worker.execute(TransferMode.SAFE, {
+                    counter to so
+                }) {(count, lock) ->
+                    repeat(1000) {
```
elizarov: `1000` needs its own constant, since it is used below in `COUNT * 1000` formula for expected value of counter

atomicfu/src/nativeTest/kotlin/kotlinx/atomicfu/test/SynchronizedTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import kotlinx.atomicfu.*
+import kotlin.native.concurrent.*
+import kotlin.native.concurrent.AtomicInt
+import kotlin.test.*
+
+class SynchronizedTest {
+
+    private val iterations = 1000
+    private val COUNT = 20
```
elizarov: Please, give it a more descriptive name. It is count of what? `nWorkers` would be a better name, for example.

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/Synchronized.kt
```diff
@@ -0,0 +1,204 @@
...
+                    if (lock.compareAndSet(state, thinLock)) return
+                }
+                THIN -> {
+                    if (currentThreadId == state.ownerThreadId) {
+                        // reentrant lock
+                        val thinNested = LockState(THIN, state.nestedLocks + 1, state.waiters, currentThreadId)
+                        if (lock.compareAndSet(state, thinNested)) return
+                    } else {
+                        // another thread is trying to take this lock -> allocate native mutex
+                        val mutex = mutexPool.allocate(locked = true)
```
elizarov: It don't see why you need to pass `locked = true` to `mutexPool.allocate`. You can just call `mutexPool.allocate()` and call `mutex.lock()` here right after `allocate()`. Moreover, you don't have any other calls to `mutexPool.allocate()`, so this whole boolean is an extra code anyway.

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/Synchronized.kt
```diff
@@ -0,0 +1,204 @@
...
+                }
+                else -> error("It is not possible to unlock the mutex that is not obtained")
+            }
+        }
+    }
+
+    private fun tryLockAfterResume(threadId: pthread_t) {
+        while (true) {
+            val state = lock.value
+            if (state.ownerThreadId != null) continue
```
elizarov: Can it ever be equal `null` here? What `continue` is supposed to achieve here?

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/Synchronized.kt
```diff
@@ -0,0 +1,204 @@
...
+            val mutex = state.mutex!!
+            val newState = if (state.waiters == 0) // deflate
```
elizarov: How can this `if (state.waiters == 0)` be ever true here? _This_ locker had incremented the waiters count so it looks like `assert(state.waiter >= 1)` can be safely added here.

elizarov:
Note: It does actually pass stress-test if I make it compile on MacOS by changing `pthread_self()!!`. The algorithm is broken.

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/Synchronized.kt
SokolovaMaria: When some thread releases the lock, it sets `ownerThreadId` field to null, decrements the number of waiters and wakes up one of the waiting threads. So, here when the resumed thread tries to obtain the lock in `tryLockAfterResume` it is not accounted in the `waiters` and it deflates the lock if there are no more waiters.

PR #100, opened 2019-11-06
ReentrantLock support added

SokolovaMaria:
`ReentrantLock` added
JVM: typealias on `juc.locks.ReentrantLock`
JS: no-op lock
Native: implementation with `LockState`

### DIFF
--- a/atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/locks/Synchronized.common.kt
+++ b/atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/locks/Synchronized.common.kt
@@ -2,5 +2,12 @@ package kotlinx.atomicfu.locks
 
 public expect open class SynchronizedObject() // marker abstract class
 
-public expect inline fun <T> synchronized(lock: SynchronizedObject, block: () -> T): T
+public expect fun reentrantLock(): ReentrantLock
 
+public expect class ReentrantLock {
+    fun lock(): Unit
+    fun tryLock(): Boolean
+    fun unlock(): Unit
+}
+
+public expect inline fun <T> synchronized(lock: SynchronizedObject, block: () -> T): T
--- a/atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/locks/Synchronized.common.kt
+++ b/atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/locks/Synchronized.common.kt
@@ -10,4 +10,6 @@ public expect class ReentrantLock {
     fun unlock(): Unit
 }
 
+public expect inline fun <T> ReentrantLock.withLock(block: () -> T): T
+
 public expect inline fun <T> synchronized(lock: SynchronizedObject, block: () -> T): T
--- a/atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
+++ b/atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
@@ -5,11 +5,11 @@ import interop.*
 import kotlinx.cinterop.*
 import kotlin.native.concurrent.*
 import kotlin.native.internal.NativePtr
-import kotlinx.atomicfu.locks.Status.*
+import kotlinx.atomicfu.locks.SynchronizedObject.Status.*
 
 public actual open class SynchronizedObject {
 
-    private val lock = AtomicReference(LockState(UNLOCKED, 0, 0))
+    protected val lock = AtomicReference(LockState(UNLOCKED, 0, 0))
 
     public fun lock() {
         val currentThreadId = pthread_self()!!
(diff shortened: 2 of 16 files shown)


atomicfu/src/jsMain/kotlin/kotlinx/atomicfu/Synchronized.kt
```diff
@@ -2,5 +2,10 @@ package kotlinx.atomicfu
 
 public actual typealias SynchronizedObject = Any
 
+public actual class ReentrantLock {
+    actual fun tryLock() = true
```
elizarov: Should be `inline`, need to test if it works.

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/Synchronized.kt
```diff
@@ -65,6 +65,27 @@ public actual open class SynchronizedObject {
...
+            val thinLock = LockState(THIN, 1, 0, currentThreadId)
+            if (lock.compareAndSet(state, thinLock)) {
+                return true
+            }
+        } else {
+            if (currentThreadId == state.ownerThreadId) {
+                check(state.status == THIN)
+                check(state.waiters == 0)
+                val nestedLock = LockState(THIN, state.nestedLocks + 1, state.waiters, currentThreadId)
+                if (lock.compareAndSet(state, nestedLock)) {
```
elizarov: It must retry on CAS failure. It cannot return `false` simply due to interference from other thread. It can only return `false` only when the lock is held by another thread.

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/Synchronized.kt
```diff
@@ -65,6 +65,30 @@ public actual open class SynchronizedObject {
...
+                THIN -> {
+                    if (currentThreadId == state.ownerThreadId) {
+                        val nestedLock = LockState(THIN, state.nestedLocks + 1, state.waiters, currentThreadId)
+                        if (lock.compareAndSet(state, nestedLock))
+                            return true
+                    } else {
+                        return false
+                    }
+                }
+                FAT -> return false
```
elizarov: What if this thread holds the lock? Please, also add a test for this case.

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Synchronized.common.kt
```diff
@@ -2,5 +2,12 @@ package kotlinx.atomicfu
 
 public expect open class SynchronizedObject() // marker abstract class
 
-public expect inline fun <T> synchronized(lock: SynchronizedObject, block: () -> T): T
+public expect fun reentrantLock(): ReentrantLock
```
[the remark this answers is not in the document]
SokolovaMaria: But there will be an overload: ReentrantLock constructor and this function.
And this name is consistent  with `atomic` constructor functions.

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/locks/Synchronized.common.kt
```diff
@@ -2,5 +2,14 @@ package kotlinx.atomicfu.locks
 
 public expect open class SynchronizedObject() // marker abstract class
 
-public expect inline fun <T> synchronized(lock: SynchronizedObject, block: () -> T): T
+public expect fun reentrantLock(): ReentrantLock
```
elizarov: Constructor functions a written with a capital letter: `reentrantLock()` -> `ReentrantLock`

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -129,12 +149,25 @@ public actual open class SynchronizedObject {
...
+
     private fun CPointer<mutex_node_t>.lock() = lock(this.pointed.mutex)
 
     private fun CPointer<mutex_node_t>.unlock() = unlock(this.pointed.mutex)
 }
 
-private enum class Status { UNLOCKED, THIN, FAT }
+public actual fun reentrantLock() = ReentrantLock()
+
+public actual class ReentrantLock : SynchronizedObject()
```
elizarov: You don't need a separate class on Kotlin/Native. Use `typealias ReentrantLock = SynchronizedObject`.

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Synchronized.common.kt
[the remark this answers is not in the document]
elizarov: This not is not consistent with existing JVM code which creates locks using `ReentrantLock()`.  You would not have overload problems if you use typealises.

PR #113, opened 2019-11-28
Fix native targets build in idea

PavelPunegov:
Fix IDEA build: fix setting native compilations


PavelPunegov:
> Could you please elaborate what's the point of this change?
> Everything works form IDEA, native is properly compiled from Gradle cmd, what does this patch fixes?

This is to fix an issue with building the Atomicfu from IDEA, and not from command-line. In IDEA it doesn't build cinterop and fails when tries to compile native code that depends on cinterop.

PR #114, opened 2019-12-11
Tracing atomic operations

SokolovaMaria:


atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Pass it to `atomic` constructor to automatically trace all modifications of the corresponding field:
+ *
+ * ```
+ * val a = atomic(initialValue, trace)
+ * ```
+ */
+@Suppress("FunctionName")
+@JsName("atomicfu\$Trace\$")
+public fun Trace(size: Int = 32, format: (Int, String) -> String = { index, text -> "$index: $text" }): Trace =
```
elizarov: This function must be `@InlineOnly` on JVM (otherwise its debug metadata pollutes the resulting class files and `AtomicfuBytecodeTest` would not pass)

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public object None : Trace()
+}
+
+private class TraceImpl(size: Int, val format: (Int, String) -> String) : Trace() {
+    init { require(size >= 1) }
+    private val size = ((size shl 1) - 1).takeHighestOneBit() // next power of 2
+    private val mask = this.size - 1
+    private val trace = arrayOfNulls<String>(this.size)
+    private var index = 0
```
elizarov: `index` must be `AtomicInteger` on JVM. The `Trace(...)` constructor function must be `execpt fun` and this `TraceImpl` should be moved to JVM-only module. The JS and Native implementation shall can use `Trace.None`.

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
```diff
@@ -20,7 +20,7 @@ import java.util.concurrent.atomic.AtomicReferenceFieldUpdater
  * private val f = atomic<Type>(initial)
  * ```
  */
-public actual fun <T> atomic(initial: T): AtomicRef<T> = AtomicRef<T>(initial)
+public actual fun <T> atomic(initial: T, trace: Trace): AtomicRef<T> = AtomicRef<T>(initial)
```
elizarov: The tracing on JVM is not actually implemented yet. It should pass `trace` to `AtomicRef`, which would use to append messages to trace on every change of value. Same for `AtomicInt` and `AtomicLong`.

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,63 @@
...
+        append(text())
+    }
+
+    /**
+     * NOP tracing.
+     */
+    public object None : Trace()
+}
+
+public class TraceImpl(size: Int, val format: (Int, String) -> String) : Trace() {
```
elizarov: It should be`private`.  Also, it should not be in `commonMain` sourceset --- it is used only in `jvmMain`.

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,63 @@
...
+    init { require(size >= 1) }
+    private val size = ((size shl 1) - 1).takeHighestOneBit() // next power of 2
+    private val mask = this.size - 1
+    private val trace = arrayOfNulls<String>(this.size)
+    private var index = atomic(0)
```
elizarov: The use of `atomic` here looks very suspcisious. It will not be processed by transformer and will be an efficient wrapper.

atomicfu/src/jsMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,6 @@
+package kotlinx.atomicfu
+
+import kotlin.js.JsName
+
+@JsName("atomicfu\$Trace\$")
+public actual fun Trace (size: Int, format: (Int, String) -> String): Trace = Trace.None
```
elizarov: Style: Extra space between `Trace` and `(`.

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -0,0 +1,26 @@
...
+    private val mask = this.size - 1
+    private val trace = arrayOfNulls<String>(this.size)
+    private val index = AtomicInteger(0)
+
+    override fun append(text: String) {
+        val i = index.getAndIncrement()
+        trace[i and mask] = format(index.get(), text)
+    }
+
+    override fun toString(): String = trace.filterNotNull().joinToString()
```
elizarov: This is Ok for tests, but no so good for real use-code that might want to dump trace to the console in case when something went wrong. It would be hard to parse when trace index had rolled over, so I'd suggest the following improvements:
* Start `toString` from the current `index` in array (then wrap around the end), so that all messages get output in their chronological order. 
* Use `/n` (LF) as a separator. This way it is going to look nice if it gets printed to any kind of log.

atomicfu/src/jvmTest/kotlin/kotlinx/atomicfu/test/TraceToStringTest.kt
```diff
@@ -0,0 +1,22 @@
...
+import kotlinx.atomicfu.atomic
+import org.junit.Test
+import kotlin.test.assertEquals
+
+class TraceToStringTest {
+    private val aTrace = Trace { i, text -> "[$i: $text]" }
+    private val a = atomic(0, aTrace)
+
+    @Test
+    fun testTraceToStringAtomicInt() {
```
elizarov: 👍 Very good. You should also add a simple test for manual `aTrace { ... }` calls.

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
```diff
@@ -72,6 +72,7 @@ public actual class AtomicRef<T> internal constructor(value: T) {
         set(value) {
             interceptor.beforeUpdate(this)
             field = value
+            trace { "value.set($value)" }
```
elizarov: This line adds a lot of string allocations to any atomic-using code even when trace is not used (which is default). I'd suggest to change this line to:
```            
if (trace !== TraceBase.None) trace { "value.set($value)" }
```
and do the same in all other AtomicXxx methods.

elizarov:
Great job. It is much better now. However, I still see the following unused entry in LVT of `TraceUseUpdate.test`:
```
LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0       0     4 $i$a$-invoke-TraceUseTest$update$1   I  <---------- THIS
            0      11     0  this   Lbytecode_test/TraceUseTest;
            0      11     1     x   I
```
It should be removed by the code that remove `trace { ... }` calls.

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -22,5 +22,5 @@ private class TraceImpl(size: Int, val format: (Int, String) -> String) : TraceB
         trace[i and mask] = format(index.get(), text)
     }
 
-    override fun toString(): String = trace.filterNotNull().joinToString()
+    override fun toString(): String = trace.slice(0 .. ((index.get() - 1) and mask)).joinToString("\n")
```
elizarov: That does not work when index is wrapped to zero. It should go from index all around the array back to index (filtering out nulls).

elizarov:
I've fixed a bunch of recent issues, but the JS transformer needs to be corrected to accommodate the latest changes.

elizarov:
I've squashed into a single commit for ease of rebasing and rebased onto the latest `develop` branch.

SokolovaMaria:
- `Trace` is now completely removed from JS.
- `TraceFormat` references removed on JVM: non-default `Trace` constructor with custom TraceFormat supported, unused references from  `trace {  }` invocations are removed from LVT.

Though `SourceDebugInfo` attribute still contains `TraceFormat` reference. Should we manually remove this reference from `debug` parameter passed to `fun visitSource(source: String?, debug: String?)` ?

elizarov (approved):
👏 Very good. I've just added some minor changes to build files so that it builds when merged with the `master` branch (which has switched to Kotlin 1.4). Let's wait until we solve build problems on Windows teamcity agents, then merge.

PR #117, opened 2019-12-26
Add 32 and 64 bit Linux ARM build targets

ionspin:
Background:
I tried compiling a multiplatform binary that included kotlin coroutines for Linux ARM 64 build target when I found out that there is no published artifact for that target. Then I found that coroutines depend on atomicfu, so I'm starting from here.

I've ran tests with ./gradlew check, but I haven't done any deep dive into the code, so I might not be aware of all the consequences of including these build targets.

**Summary:**

Add 32 and 64 bit Linux ARM targets so the library can be used on ARM devices running Linux such as Raspberry PI.


elizarov:
We are holding off these additions for now because we currently lack the infrastructure to test our libraries on these platforms and we feel reluctant to release artifacts that did not go through any kind of testing.

PR #125, opened 2020-02-27
Enable JS IR backend

skuzmich:
Use new K/JS Gradle plugin for tests and publishing
Migrate to built-in node.js and Chrome test runners

### DIFF
--- a/atomicfu/build.gradle
+++ b/atomicfu/build.gradle
@@ -148,6 +148,7 @@ repositories.whenObjectAdded {
     }
 }
 
+def transformedJsFile = compileTestKotlinJs.kotlinOptions.outputFile
 compileTestKotlinJs {
     kotlinOptions {
         // NOTE: Module base-name must be equal to the package name declared in package.json
@@ -155,8 +156,7 @@ compileTestKotlinJs {
         outputFile = new File(outputFile.parent, baseName + ".js")
     }
 }
-def originalJsFile = "${rootProject.buildDir}/js/packages/kotlinx.atomicfu-atomicfu-test/kotlin/kotlinx-atomicfu.js"
-def transformedJsFile = "${rootProject.buildDir}/js/packages/kotlinx.atomicfu-atomicfu-test/kotlin/kotlinx.atomicfu-atomicfu-test.js"
+def originalJsFile = compileTestKotlinJs.kotlinOptions.outputFile
 
 task transformJS(type: JavaExec, dependsOn: [compileTestKotlinJs]) {
     main = "kotlinx.atomicfu.transformer.AtomicFUTransformerJSKt"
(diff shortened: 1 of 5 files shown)


gradle/publish-bintray.gradle
```diff
@@ -74,7 +74,7 @@ publishing {
 
         // disable metadata everywhere, but in native modules
```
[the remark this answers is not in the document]
skuzmich: Updated

skuzmich:
Updated on latest `develop`

atomicfu/build.gradle
```diff
@@ -156,7 +166,13 @@ task transformJS(type: JavaExec, dependsOn: [compileTestKotlinJs]) {
     outputs.file(transformedJsFile)
 }
 
-testMochaNode.dependsOn transformJS
```
qwwdfsad: mocha tests are now disabled from `build` task?

skuzmich: Custom js tests are disabled. Built-in tests 

* build
  * legacyjsTest 
    * legacyjsNodeTest
    * legacyjsBrowserTest
  * jsIrTest 
    * jsIrNodeTest
    * jsIrBrowserTest

are enabled

skuzmich:
Fixed issue with test file passes on TeamCity build. Gradle projcet has no name specified (I assumed it was `kotlinx.atomicfu`) so generated js files had prefix with tmp directory name like `88b0986a7186d029`.

Disabled borwser tests since TeamCity agents don't have chrome.

TeamCity build is green now.

gradle.properties
```diff
@@ -17,4 +17,6 @@ mocha_teamcity_reporter_version=2.2.2
 source_map_support_version=0.5.3
 
 kotlin.incremental.multiplatform=true
-kotlin.native.ignoreDisabledTargets=true
\ No newline at end of file
+kotlin.native.ignoreDisabledTargets=true
+
+kotlin.js.compiler=both
```
qwwdfsad: Could you please elaborate on the specifics of this change? It's okay to do it in person, not here, whatever suits you.

Do I get it right: with 1.4-M1 the library is built twice (using old BE and IR BE) and published as an aggregated artifact with a corresponding metadata. And for Kotlin 1.4, Kotlin plugin knows how to pick the right artifact (old or a new one) using the published metadata?

skuzmich: You are right. We build and publish both `.jar` and `.klib` with the same name and metadata to help consumers choose which one to use

skuzmich:
> According to the teamcity report, 66 tests were ran on your branch, while on the rest of the branches its 117.

Known issue. MPP tests are squashed into a single test with multiple runs.

For example, [this test used to have 7 runs](https://teamcity.jetbrains.com/viewLog.html?buildId=2792897&buildTypeId=KotlinTools_KotlinxAtomicfu_BuildWindows&tab=testsInfo)

But now [it has 8](https://teamcity.jetbrains.com/viewLog.html?buildTypeId=KotlinTools_KotlinxAtomicfu_BuildLinux&buildId=2790739&tab=testsInfo&pager.currentPage=1)

PR #127, opened 2020-03-12
Support array initialization generated by JVM_IR

ilmirus:
Old BE generates array initialization without ASTOREs,
like
  ICONST_1
  ANEWARRAY L/java/lang/String;
  DUP
  ICONST_0
  LDC "A"
  AASTORE

But JVM_IR generates array initialization like this:
  ICONST_1
  ANEWARRAY L/java/lang/String;
  ASTORE_0
  ALOAD_0
  ICONST_0
  LDC "A"
  AASTORE
  ALOAD_0

In other words, it empties the stack after each element.
Thus, we cannot just assume, that empty stack means initialization start,
instead, we need to find array ANEWARRAY or NEWARRAY
(thankfully, JVM_IR does not generate MULTIANEWARRAY)
and constant before it.


ilmirus:
By the way, I wonder, why don't we use SourceInterpreter, which will point to ANEWARRAY in this case, and initialization start in others?

PR #132, opened 2020-04-16
Ddol/update interop klib publishing 14 m2

ddolovov:


atomicfu-gradle-plugin/src/test/kotlin/kotlinx/atomicfu/plugin/gradle/Project.kt
```diff
@@ -86,6 +86,7 @@ class Project(val projectDir: File) {
                     maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
                     maven { url 'https://dl.bintray.com/kotlin/kotlin-dev' }
                     maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
+
```
qwwdfsad, line 89: redundant blank line in an otherwise untouched file. Please avoid that

build.gradle
```diff
@@ -69,6 +70,7 @@ allprojects {
             }
         }
         maven { url "https://kotlin.bintray.com/kotlinx" }
+        maven { url kotlinCompilerRepo }
```
qwwdfsad, line 73: please document this line and enable it conditionally only for snapshot builds

qwwdfsad: And please elaborate on why teamcity coordinates are required in order to build afu on Kotlin master

ddolovov: That's not my commit. Let's create a separate clean review.

PR #134, opened 2020-04-17
Fixes for Kotlin 1.4-M2

h0tk3y:


h0tk3y:
Note that this effectively drops the Maven module `atomicfu-native`. If any consumer specified a dependency on it explicitly, that will no longer work. I've updated the Gradle plugin so that it adds the appropriate dependency. I can also check the other kotlinx libraries for any explicit dependencies and get  back to you. Until then, you can cherry-pick jus the commit with cinterop publishing adjustments.

PR #143, opened 2020-06-16
Atomicfu Js/Ir compiler plugin support in the gradle plugin

SokolovaMaria:


elizarov:
Is this PR still active? It does not build.

PR #146, opened 2020-07-28
- add Windows 32bit support

danielgruber:
- Build and Tests work on my local machine


qwwdfsad:
Alas, we are not _yet_ ready to enable new Native targets.
We are working on the list with full Kotlin/Native-wide list of various targets support (debugging support/unit testing support/whether we test our libraries on a specific target and so on) and as soon as it's ready, we'll enable _all_ existing targets, referencing a giving guarantees

PR #153, opened 2020-10-11
FU and VH fields marked as ACC_SYNTHETIC

### LINKED ISSUE #144: Mark generated $FU fields as ACC_SYNTETIC
That will ensure that they cannot be reflected in Java, which is logical because they are essentially a private implementation detail not originally present in the source code. See discussion in https://youtrack.jetbrains.com/issue/KT-40328

SokolovaMaria:
Fixes #144

### DIFF
--- a/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
+++ b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
@@ -421,7 +421,7 @@ class AtomicFUTransformer(
                 val fieldId = FieldId(className, name, desc)
                 val f = fields[fieldId]!!
                 val protection = when {
-                    // reference to wrapper class (primitive atomics) or reference to to j.u.c.a.Atomic*Array (atomic array)
+                    // reference to wrapper class (primitive atomics) or reference to j.u.c.a.Atomic*Array (atomic array)
                     f.isStatic && !vh -> ACC_STATIC or ACC_FINAL or ACC_SYNTHETIC
                     // primitive type field
                     f.isStatic && vh -> ACC_STATIC or ACC_SYNTHETIC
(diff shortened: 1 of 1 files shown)


atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -446,10 +446,10 @@ class AtomicFUTransformer(
                 }
                 if (vh) {
                     // VarHandle is needed for all array element accesses and for regular fields with atomic ops
-                    if (f.hasAtomicOps || f.isArray) vhField(protection, f)
+                    if (f.hasAtomicOps || f.isArray) vhField(protection or ACC_SYNTHETIC, f)
                 } else {
                     // FieldUpdater is not needed for arrays (they use AtomicArrays)
-                    if (f.hasAtomicOps && !f.isArray) fuField(protection, f)
+                    if (f.hasAtomicOps && !f.isArray) fuField(protection or ACC_SYNTHETIC, f)
```
elizarov, line 452: There's already ACC_SYNTHETIC set in `protection` at lines 425-428 with the only exception of a private field (where it is not needed). https://youtrack.jetbrains.com/issue/KT-40328 should have been caused by something else. It needs better investigation and the corresponding test-case.

elizarov (changes requested):

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
SokolovaMaria: Actually, the `protection` for package-private fields didn't include `ACC_SYNTHETIC`. 

And this was the case from the issue :point_down: , the inner class `B` caused an accessor in the bytecode for the private field `state`, so the corresponding `FU` field was marked package-private (non-synthetic). 
```
class A {
    private val state = atomic(0)

    inner class B {
        ...
    }
} 
```
Though as there was no actual access to the `state` from the inner class `JVM_IR` didn't not generate an accessor, so the field was private there.

SokolovaMaria: The corresponding case on access modifiers is provided in this PR #154

elizarov (changes requested):
Can you then, please, merge changes from this PR into #154, so it would be easier to review with all the tests.

SokolovaMaria:
Changes from this PR merged to #154

PR #154, opened 2020-10-11
Access modifiers and naming of *RefVolatile classes fixed.

SokolovaMaria:
Access modifiers fixed for: 

- `*RefVolatile` classes
- volatile primitive type fields inside `*RefVolatle` classes
- synthetic `FU` and `*RefVolatile` fields

### DIFF
--- a/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
+++ b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
@@ -1245,14 +1245,10 @@ class AtomicFUTransformer(
 
         // generates a ref class with volatile field of primitive type inside
         private fun generateRefVolatileClass(f: FieldInfo, arg: Type) {
-            val visibility = when {
-                f.hasExternalAccess -> ACC_PUBLIC
-                f.accessors.isEmpty() -> ACC_PRIVATE
-                else -> 0
-            }
             if (analyzePhase2) return // nop
             val cw = ClassWriter(0)
-            cw.visit(V1_6, visibility or ACC_SYNTHETIC, f.refVolatileClassName, null, "java/lang/Object", null)
+            val classVisibility = if (f.hasExternalAccess) ACC_PUBLIC else 0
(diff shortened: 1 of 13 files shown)


elizarov (changes requested):
Let's use this as an opportunity to improve the testing infrastructure of atomicfu. We need to have tests for those kinds of naming and access-modifer changes. I suggest to write a small test framework that takes a class and, using reflection, prints all field declarations in this class in a Java-like human-readable form with their names, access modifiers, and type. This way you can then write tests with `assertEquals`, to test this kind of changes.

This change really needs this kind of test to see what's going on here.

elizarov (changes requested):
It does not pass tests.

SokolovaMaria:
Changes from PR #153 merged:
- marking package-private fields with `ACC_SYNTHETIC` flag + reflection test

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -1242,9 +1245,14 @@ class AtomicFUTransformer(
 
         // generates a ref class with volatile field of primitive type inside
         private fun generateRefVolatileClass(f: FieldInfo, arg: Type) {
+            val visibility = when {
+                f.hasExternalAccess -> ACC_PUBLIC
+                f.accessors.isEmpty() -> ACC_PRIVATE
```
elizarov, line 1250: This is the visibility of a _class_ and class cannot be `PRIVATE`. Please, add a test for this case. You can get rid of accessors by using `@JvmField` annotation on a top-level val. Class visibility will need its own code (separate from field visibility)

Also, this code should be placed after `if (analyzePhase2) return // nop` line below.

elizarov (changes requested):

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
SokolovaMaria: Fixed this added `*RefVolatile` class modifiers check for all static field reflection tests.

elizarov (approved):
🚀 Great work

PR #155, opened 2020-10-20
Atomic*Array.size method introduced

### LINKED ISSUE #59: No size in atomic arrays
No size getter in `AtomicArray` and `array` field is private:
```
public class AtomicArray<T> internal constructor(size: Int) {
    private val array = Array(size) { atomic<T?>(null) }

    @JsName("get\$atomicfu")
    public operator fun get(index: Int): AtomicRef<T?> = array[index]
}
```

### LINKED ISSUE #149: Provide AtomicArray.length method
Currently, the only method atomic array provides is `get`, making it very hard to e.g. iterate over all its content of the array, get its length and so on, it would be nice to fix that

SokolovaMaria:
Fixes #149
Fixes #59

### DIFF
--- a/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
+++ b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
@@ -63,10 +63,8 @@ private val ARRAY_ELEMENT_TYPE: Map<Type, Int> = mapOf(
 private val AFU_TYPES: Map<Type, TypeInfo> = AFU_CLASSES.mapKeys { getObjectType(it.key) }
 
 private val METHOD_HANDLES = "$JLI_PKG/MethodHandles"
-private val METHOD_HANDLE = "$JLI_PKG/MethodHandle"
 private val LOOKUP = "$METHOD_HANDLES\$Lookup"
 private val VH_TYPE = getObjectType("$JLI_PKG/VarHandle")
-private val METHOD_HANDLE_TYPE = getObjectType(METHOD_HANDLE)
 
 private val STRING_TYPE = getObjectType("java/lang/String")
 private val CLASS_TYPE = getObjectType("java/lang/Class")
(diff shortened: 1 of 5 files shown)


atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -466,6 +470,7 @@ class AtomicFUTransformer(
         // Generates static VarHandle field
         private fun vhField(protection: Int, f: FieldInfo) {
             super.visitField(protection or ACC_FINAL or ACC_STATIC, f.fuName, VH_TYPE.descriptor, null, null)
+            super.visitField(protection or ACC_FINAL or ACC_STATIC, f.arrayLengthMethodHandleName, METHOD_HANDLE_TYPE.descriptor, null, null)
```
elizarov, line 473: I don't really see why it is even needed. We can directly access the underlying array's length without a method handle, cannot we?

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
```diff
@@ -718,45 +732,70 @@ class AtomicFUTransformer(
...
                 }
             }
+            if (f.isArray && iv.name == GET_SIZE) {
+                if (!vh) {
+                    // map to j.u.c.a.Atomic*Array length()
+                    iv.owner = descToName(f.fuType.descriptor)
+                    iv.name = "length"
+                } else {
+                    // map to MethodHandles.arrayLength
+                    val getStaticMH = FieldInsnNode(
```
elizarov, line 742: We can just use `arraylength` bytecode here. There's no need to use a MethodHandle for it.

elizarov (changes requested):

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformer.kt
SokolovaMaria: Got it, no need to use MethodHandle for this. I've erased mapping to MethodHandles.arrayLength and generating the static field.

elizarov (commented):
Great

PR #157, opened 2020-11-02
Add JVM IR target in build_snapshot_train mode

ALikhachev:


ALikhachev:
During private discussion we've decided to try another way. I'll open new PR.

PR #167, opened 2020-12-03
Version 0.15.0

SokolovaMaria:


CHANGES.md
```diff
@@ -1,5 +1,13 @@
 # Change log for kotlinx.atomicfu
 
+# Version 0.15.0
+
+* Tracing atomic operations
```
qwwdfsad, line 5: It's worth linking the corresponding issues/PR's as we do in coroutines: https://github.com/Kotlin/kotlinx.coroutines/blob/master/CHANGES.md

SokolovaMaria: Done

PR #169, opened 2020-12-11
Removing reference to TraceBase class from JS

SokolovaMaria:
JS transformer is fixed to remove the reference to `kotlinx.atomicfu.TraceBase` class


elizarov (approved):
👍 Loogs good.

PR #170, opened 2020-12-11
Binary compatibility

SokolovaMaria:
The `Trace` parameter with a default value, added to atomic factory functions, broke binary compatibility in`0.15.0` 

For now, the fix is to return old declarations of atomic factory functions with one parameter.

### DIFF
--- a/atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
+++ b/atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
@@ -12,66 +12,94 @@ import kotlinx.atomicfu.TraceBase.None
 import kotlin.reflect.KProperty
 
 /**
+ * Creates atomic reference with a given [initial] value and a [trace] object to trace modifications of the value.
+ *
+ * It can only be used to initialize a private or internal read-only property, like this:
+ *
+ * ```
+ * private val f = atomic<Type>(initial, trace)
+ * ```
+ */
+public expect fun <T> atomic(initial: T, trace: TraceBase = None): AtomicRef<T>
+
+// Binary compatibility with IR, should be removed with Kotlin 1.5 release
+/**
  * Creates atomic reference with a given [initial] value.
  *
- * It can only be used in initialize of private read-only property, like this:
+ * It can only be used to initialize a private or internal read-only property, like this:
  *
  * ```
  * private val f = atomic<Type>(initial)
  * ```
  */
-public expect fun <T> atomic(initial: T, trace: TraceBase = None): AtomicRef<T>
-
-/**
- * Binary compatibility with IR, should be removed with Kotlin 1.5 release
- */
 public expect fun <T> atomic(initial: T): AtomicRef<T>
 
 /**
+ * Creates atomic [Int] with a given [initial] value and a [trace] object to trace modifications of the value.
+ *
+ * It can only be used to initialize a private or internal read-only property, like this:
+ *
+ * ```
+ * private val f = atomic(initialInt, trace)
+ * ```
+ */
+public expect fun atomic(initial: Int, trace: TraceBase = None): AtomicInt
+
+// Binary compatibility with IR, should be removed with Kotlin 1.5 release
+/**
  * Creates atomic [Int] with a given [initial] value.
  *
- * It can only be used in initialize of private read-only property, like this:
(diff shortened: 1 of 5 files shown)


atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
```diff
@@ -23,6 +23,9 @@ import kotlin.reflect.KProperty0
  */
 public expect fun <T> atomic(initial: T, trace: TraceBase = None): AtomicRef<T>
 
+// Binary compatibility
```
qwwdfsad, line 26: Please drop a comment "Binary compatibility with IR, should be removed with Kotlin 1.5 release"

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
```diff
@@ -23,6 +23,9 @@ import kotlin.reflect.KProperty0
...
+public expect fun <T> atomic(initial: T): AtomicRef<T>
```
qwwdfsad, line 27: Now, these functions have missing javadoc, should be fixed

qwwdfsad (changes requested):

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
qwwdfsad: Not what I've meant :)

This comment is required _for us_ in order to maintain the library. It should be one-liner or `/* */` block in the start of the file.

Javadoc is for library users. It probably should say `Creates atomic reference with a given [initial] value.`.
And Javadoc for `atomic(initial: T, trace: TraceBase = None)` should also explain what trace parameter is

qwwdfsad (changes requested):

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
```diff
@@ -10,51 +10,98 @@ import kotlin.js.JsName
...
  *
  * ```
  * private val f = atomic<Type>(initial)
  * ```
  */
-public expect fun <T> atomic(initial: T, trace: TraceBase = None): AtomicRef<T>
+public expect fun <T> atomic(initial: T): AtomicRef<T>
 
+/**
+ * Creates atomic [Int] with a given [initial] value and a [trace] object to trace modifications of the value.
```
qwwdfsad, line 38: I'd suggest adding a link to the trace so it will be easier to understand what trace is:
`[trace modifications][Trace]`

atomicfu/src/jsMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
```diff
@@ -7,20 +7,28 @@
 package kotlinx.atomicfu
 
 import kotlin.reflect.KProperty
-import kotlin.reflect.KProperty0
+import kotlinx.atomicfu.TraceBase.None
 
 @JsName("atomic\$ref\$")
 public actual fun <T> atomic(initial: T, trace: TraceBase): AtomicRef<T> = AtomicRef<T>(initial)
 
+public actual inline fun <T> atomic(initial: T): AtomicRef<T> = atomic(initial, None)
```
qwwdfsad, line 15: Do we need `JsName` on these overloads as well?

qwwdfsad (changes requested):

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
qwwdfsad: Also, while we are here, `traceFormatDefault` doc states "the default **trade** string formatter". Probably meant to be "trace"

atomicfu/src/jsMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
SokolovaMaria: No, it is inlined and the overload with the mangled name is called

qwwdfsad (approved):
Great job! Let's wait for the build to pass (bintray seems to fluctuate) and merge it

PR #172, opened 2020-12-16
Debug trace enhancements

### LINKED ISSUE #162: Provide multiline append in atomic tracing
During debugging of tricky data-races, it is often desirable to log more than one statement successively.

While it is possible to do so by concatenating, e.g. by replacing
```
trace.append(state1)
trace.append(state2)
```
with
```
trace.append(state1 + "\n" + state2)
```

it creates unnecessary garbage. While it's okay for diagnostic purposes, it's not really acceptable for catching data-races, because such memory pressure typically changes timings and data-races stops reproducing.

It would be nice to have `append` overloads for 2-3-4 arguments (not varargs!) that reserve slots in the trace atomically.

### LINKED ISSUE #163: Provide trace implementation that accept and stores Any instead of the String
It would be nice if the default implementation (or any specialized one) could store and accept `Any` type, postponing the formatting to the moment of `toString` invocation.

It allows storing threads, enums and integer constants without memory pressure that often obscures data-race hunting. See rationale in #162

SokolovaMaria:
Fixes #162
Fixes #163

### DIFF
--- a/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
+++ b/atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
@@ -13,16 +13,16 @@ import java.util.regex.*
 
 private const val ATOMIC_CONSTRUCTOR = """(atomic\$(ref|int|long|boolean)\$|Atomic(Ref|Int|Long|Boolean))"""
 private const val ATOMIC_CONSTRUCTOR_BINARY_COMPATIBILITY = """(atomic\$(ref|int|long|boolean)\$1)""" // mangled names for declarations left for binary compatibility
-private const val ATOMIC_ARRAY_CONSTRUCTOR = """Atomic(Ref|Int|Long|Boolean)Array\$(ref|int|long|boolean|ofNulls)"""
+private const val ATOMIC_ARRAY_CONSTRUCTOR = """(atomicfu)\$(Atomic(Ref|Int|Long|Boolean)Array)\$(ref|int|long|boolean|ofNulls)"""
 private const val MANGLED_VALUE_PROP = "kotlinx\$atomicfu\$value"
 
-private const val TRACE_CONSTRUCTOR = "Trace\\\$atomicfu\\\$"
-private const val TRACE_BASE_CLASS = "TraceBase\\\$atomicfu\\\$"
-private const val TRACE_APPEND = """(Trace)\$(append)\$([1234])\$(atomicfu)\$""" // [1234] is the number of arguments in the append overload
-private const val TRACE_NAMED = "Trace\\\$named\\\$atomicfu\\\$"
+private const val TRACE_CONSTRUCTOR = "atomicfu\\\$Trace"
+private const val TRACE_BASE_CLASS = "atomicfu\\\$TraceBase"
+private const val TRACE_APPEND = """(atomicfu)\$(Trace)\$(append)\$([1234])""" // [1234] is the number of arguments in the append overload
+private const val TRACE_NAMED = "atomicfu\\\$Trace\\\$named"
 private const val TRACE_FORMAT = "TraceFormat"
-private const val TRACE_FORMAT_CONSTRUCTOR = "$TRACE_FORMAT\\\$atomicfu\\\$"
-private const val TRACE_FORMAT_FORMAT = "$TRACE_FORMAT\\\$format\\\$atomicfu\\\$"
+private const val TRACE_FORMAT_CONSTRUCTOR = "atomicfu\\\$$TRACE_FORMAT"
+private const val TRACE_FORMAT_FORMAT = "atomicfu\\\$$TRACE_FORMAT\\\$format"
 
 private const val RECEIVER = """(\$(receiver)(_\d+)?)"""
 private const val SCOPE = "scope"
@@ -36,11 +36,11 @@ private const val ATOMIC_REF = "AtomicRef"
 private const val MODULE_KOTLINX_ATOMICFU = "\\\$module\\\$kotlinx_atomicfu"
 private const val ARRAY = "Array"
 private const val FILL = "fill"
-private const val GET_ELEMENT = "get\\\$atomicfu\\\$"
-private const val ARRAY_SIZE = "size\$atomicfu\$"
+private const val GET_ELEMENT = "atomicfu\\\$get"
+private const val ARRAY_SIZE = "atomicfu\$size"
 private const val LENGTH = "length"
 private const val LOCKS = "locks"
-private const val REENTRANT_LOCK_ATOMICFU_SINGLETON = "$LOCKS.reentrantLock\\\$atomicfu\\\$"
+private const val REENTRANT_LOCK_ATOMICFU_SINGLETON = "$LOCKS.atomicfu\\\$reentrantLock"
 
 
 private val MANGLE_VALUE_REGEX = Regex(".${Pattern.quote(MANGLED_VALUE_PROP)}")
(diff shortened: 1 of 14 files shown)


elizarov:
I'm not entirely sure this does what @qwwdfsad has had in mind.

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
```diff
@@ -16,7 +16,7 @@ private const val ATOMIC_ARRAY_CONSTRUCTOR = """Atomic(Ref|Int|Long|Boolean)Arra
 private const val MANGLED_VALUE_PROP = "kotlinx\$atomicfu\$value"
 
 private const val TRACE_CONSTRUCTOR = "atomicfu\\\$Trace\\\$"
-private const val TRACE_APPEND = "atomicfu\\\$Trace\\\$append\\\$"
+private const val TRACE_APPEND = """atomicfu\$(Trace)\$(append)\$(1|2|3|4)\$"""
```
qwwdfsad, line 19: Please ensure that new methods are correctly eliminated from JS

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.common.kt
```diff
@@ -60,12 +60,24 @@ public expect fun TraceBase.named(name: String): TraceBase
 public expect val traceFormatDefault: TraceFormat
 
 public open class TraceBase internal constructor() {
-    @JsName("atomicfu\$Trace\$append\$")
+    @JsName("atomicfu\$Trace\$append\$1\$")
```
qwwdfsad, line 63: Could you please elaborate on why `JsName` is required here?

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/TraceFormat.kt
```diff
@@ -18,14 +18,14 @@ public open class TraceFormat {
      * Formats trace at the given [index] with the given [text] as string.
```
qwwdfsad, line 18: Documentation and parameter name are now obsolete, please update it

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -24,41 +24,74 @@ public actual val traceFormatDefault: TraceFormat =
...
+    override fun format(index: Int, text: Any): String =
         "$index: [${Thread.currentThread().name}] $text"
 }
 
 private class NamedTrace(
     private val trace: TraceBase, 
     private val name: String
 ) : TraceBase() {
-    override fun append(text: String) = trace.append("$name.$text")
+    override fun append(arg1: Any) = trace.append("$name.$arg1")
```
qwwdfsad, line 35: it would be nice to give these parameters a more meaningful name (in the base class)

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -24,41 +24,74 @@ public actual val traceFormatDefault: TraceFormat =
...
 private class TraceImpl(size: Int, val format: TraceFormat) : TraceBase() {
     init { require(size >= 1) }
     private val size = ((size shl 1) - 1).takeHighestOneBit() // next power of 2
     private val mask = this.size - 1
-    private val trace = arrayOfNulls<String>(this.size)
+    private val trace = arrayOfNulls<Any>(this.size)
     private val index = AtomicInteger(0)
 
-    override fun append(text: String) {
+    override fun append(arg1: Any) {
```
qwwdfsad, line 55: All these methods are great, but they are internal and not available in user code, only `invoke {}` can be used from outside of atomicfu. Please fix that

qwwdfsad (changes requested):

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/Trace.kt
qwwdfsad: `event` name probably will do the trick

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
SokolovaMaria: I've added the test to check hat no references to atomicfu are left in the js output

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.common.kt
SokolovaMaria: I use `JsName` to rely on this name while erasing these declarations from the js output

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/Trace.kt
```diff
@@ -24,41 +24,74 @@ public actual val traceFormatDefault: TraceFormat =
...
+    override fun append(event1: Any, event2: Any, event3: Any) =
+            trace.append("$name.$event1", "$name.$event2", "$name.$event3")
+
+    override fun append(event1: Any, event2: Any, event3: Any, event4: Any) =
+            trace.append("$name.$event1", "$name.$event2", "$name.$event3", "$name.$event4")
+
     override fun toString(): String = trace.toString()
 }
 
 private class TraceImpl(size: Int, val format: TraceFormat) : TraceBase() {
```
qwwdfsad, line 48: `format` parameter can be private (also, will save a getter and setter)

atomicfu-transformer/src/main/kotlin/kotlinx/atomicfu/transformer/AtomicFUTransformerJS.kt
```diff
@@ -15,12 +15,12 @@ private const val ATOMIC_CONSTRUCTOR = """(atomic\$(ref|int|long|boolean)\$|Atom
 private const val ATOMIC_ARRAY_CONSTRUCTOR = """Atomic(Ref|Int|Long|Boolean)Array\$(ref|int|long|boolean|ofNulls)"""
 private const val MANGLED_VALUE_PROP = "kotlinx\$atomicfu\$value"
 
-private const val TRACE_CONSTRUCTOR = "atomicfu\\\$Trace\\\$"
-private const val TRACE_APPEND = "atomicfu\\\$Trace\\\$append\\\$"
-private const val TRACE_NAMED = "atomicfu\\\$Trace\\\$named\\\$"
+private const val TRACE_CONSTRUCTOR = "Trace\\\$atomicfu\\\$"
+private const val TRACE_APPEND = """(Trace)\$(append)\$(1|2|3|4)\$(atomicfu)\$"""
```
qwwdfsad, line 19: IDEA highlights this regexp and suggests replacing `(1|2|3|4)` with `[1234]`. Also, worth to drop a comment that this is a number of overloads

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.common.kt
```diff
@@ -60,12 +60,20 @@ public expect fun TraceBase.named(name: String): TraceBase
 public expect val traceFormatDefault: TraceFormat
 
 public open class TraceBase internal constructor() {
-    @JsName("atomicfu\$Trace\$append\$")
-    @PublishedApi
-    internal open fun append(text: String) {}
+    @JsName("Trace\$append\$1\$atomicfu\$")
+    public open fun append(event: Any) {}
```
qwwdfsad, line 64: Please fix all the implementors of this class, because they are still using `text` parameter name

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.common.kt
```diff
@@ -60,12 +60,20 @@ public expect fun TraceBase.named(name: String): TraceBase
 public expect val traceFormatDefault: TraceFormat
 
 public open class TraceBase internal constructor() {
```
qwwdfsad, line 62: This class is public and user-facing, but neither it nor its method has a documentation

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
```diff
@@ -461,11 +461,11 @@ public inline fun AtomicLong.updateAndGet(function: (Long) -> Long): Long {
 public class AtomicIntArray(size: Int) {
     private val array = Array(size) { atomic(0) }
 
-    @JsName("size\$atomicfu")
+    @JsName("size\$atomicfu\$")
```
qwwdfsad, line 464: Great job with meticulous JS methods marking!

Unfortunately, it contains a lot of not-so-readable symbols that have not so clear purpose.
It would be nice if we could have a single `.kt` file with all these names in a single place as `internal const val`
and then reference it on the use-site:
```
@JsName(JS_ARRAY_GET)
public operator fun get(index: Int): AtomicInt
```

Then we will have a single place with all the documentation and explanation of what it is and why it is written in this way. As bouns, JVM users won't see it, but curious ones will be able to study it if they want.
WDYT?

qwwdfsad (changes requested):

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/AtomicFU.common.kt
SokolovaMaria: Moved all mangled names to one file, really looks much better now

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/MangledJsNames.kt
```diff
@@ -0,0 +1,63 @@
+package kotlinx.atomicfu
+
+/**
+ * All atomicfu declarations are annotated with [@JsName][kotlin.js.JsName] to have specific names in JS output.
```
qwwdfsad, line 4: 👍 much better!

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/Trace.common.kt
```diff
@@ -23,6 +23,10 @@ import kotlin.internal.InlineOnly
  * ```
  * trace { "Doing something" }
  * ```
+ * or
+ * ```
+ * trace.append(Thread.currentThread())
+ * ```
```
qwwdfsad, line 29:
```suggestion
 * or you can do multi-append in a garbage-free manner
 * ```
 * // Before queue.send(element) invocation
 * trace.append("Adding element to the queue", element, Thread.currentThread())
 * ```
```

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/MangledJsNames.kt
```diff
@@ -0,0 +1,63 @@
...
+internal const val ATOMIC_BOOLEAN_FACTORY = "atomic\$boolean\$"
+internal const val ATOMIC_BOOLEAN_FACTORY_BINARY_COMPATIBILITY = "atomic\$boolean\$1"
+
+// Atomic value
+internal const val ATOMIC_VALUE = "kotlinx\$atomicfu\$value"
+
+// Atomic operations
+internal const val COMPARE_AND_SET = "compareAndSet\$atomicfu\$"
+internal const val GET_AND_SET = "getAndSet\$atomicfu\$"
+internal const val GET_AND_INCREMENT = "getAndIncrement\$atomicfu\$"
```
qwwdfsad, line 24: could you please elaborate on why `atomicfu` substring is inserted in the middle of the signature?

I do not really mind, it just looks a bit alien and a more familiar `atomicfu\$getAndIncrement` can be used instead

qwwdfsad (approved):
Otherwise good to go, great job here!

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/MangledJsNames.kt
SokolovaMaria: Actually, this turned out to be just a crutch for JS transformer that is no longer needed. So, I also fixed all other mangled names to have `atomicfu$` prefix for consistency.

PR #173, opened 2020-12-16
Maven central publication

qwwdfsad:


buildSrc/src/main/kotlin/Publishing.kt
```diff
@@ -0,0 +1,94 @@
...
+fun MavenPom.configureMavenCentralMetadata(project: Project) {
+    name by project.name
+    description by "AtomicFU utilities"
+    url by "https://github.com/Kotlin/kotlinx.atomicfu"
+
+    licenses {
+        license {
+            name by "The Apache Software License, Version 2.0"
+            url by "https://www.apache.org/licenses/LICENSE-2.0.txt"
+            distribution by "repo"
```
elizarov, line 29: So nice! It makes me really wonder if Kotlin language shall support some kind of C++-style `operator fun assign` to overload an assignment of read-only properties (like `name`, `url`, etc), so that it can be written simply as:

```
name = "The Apache Software License, Version 2.0"
url = "https://www.apache.org/licenses/LICENSE-2.0.txt"
distribution = "repo"
```

The downside, is, of course, is that accessing `name`, `url`, `distribution` does not give you their values, but just a `Property` reference.

elizarov (approved):
👍 Nice!

PR #174, opened 2020-12-24
Fixed running excluded tests in Idea

SokolovaMaria:
There was a problem with running tests excluded in `transformedTest*` tasks in Idea.
The following error was thrown:
```
No tests found for given includes: [**/*LFTest.*](exclude rules) [kotlinx.atomicfu.test.LockFreeQueueLFTest](filter.includeTestsMatching)
```
while these tests were executed correctly in the terminal.

Seems that Idea sets the property `failOnNoMatchingTests` to true in it's init script. 
Setting this property to false in gradle has priority over the Idea settings and solves the problem.


qwwdfsad (approved):
When adding any non-trivial configuration, please always drop a comment on why it was added in the first place.

PR #193, opened 2021-07-14
Add support for linuxArm64 and linuxArm32Hfp targets

soywiz:
Relates to: https://github.com/Kotlin/kotlinx.coroutines/issues/855


Jessewb786:
I work for Trimble and we're currently evaluating Kotlin/Native as a replacement and linuxArm32hfp is a must for us to go ahead with it. I suspect this is the case for quite a few other companies too.
For example, we need Ktor for our network communications, but it doesn't have support for the target, because coroutines doesn't have support for it, because atomicfu hasn't added support for it.
It would be greatly appreciated if this PR were given some attention.

P.S: The work you guys do is much appreciated.

PR #201, opened 2021-09-07
Add explicit module-info for JPMS compatibility

lion7:
Largely based on https://github.com/Kotlin/kotlinx.serialization/pull/1624 and the review comments there.
Note: I did have to update the ASM API version to `ASM6` for the build to succeed..


lion7:
This project consists of only 1 published runtime JAR, similar to `kotlinx.datetime`.
I moved the configuration of the module-info compilation task to an inline Gradle script, assuming that the comments of @ilya-g in https://github.com/Kotlin/kotlinx-datetime/pull/135 also apply here.

Note: I also reverted the ASM API change since it is already upgraded to `ASM9` in a previous PR. I did have to upgrade the Gralde version so I could use the Java toolchain feature.

mvicsokolova (commented):
Thanks a lot for your work! I've read the review of JPMS support for kotlinx-datetime as the structure here is pretty much the same. Looks good to me.
I've asked @ilya-g to take the final look before merge.

mvicsokolova (approved):

ilya-g (approved):

PR #204, opened 2021-09-20
Correct artifact name for instructions on use within common sourceset.

forrestpangborn:
It looks like the artifact name specified in the `README` under [the section](https://github.com/Kotlin/kotlinx.atomicfu#common) on how to include in `common` sourcesets is incorrect, has somewhat recently changed, or something failed to properly publish to Maven Central.

When I attempt to include `org.jetbrains.kotlinx:atomicfu-common:0.16.3` as a dependency in common code, resolution fails - but when I include `org.jetbrains.kotlinx:atomicfu:0.16.3`, resolution succeeds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ See [gradle.properties](gradle.properties) in AtomicFU project for its `kotlin_v
 
 ### Common
 
-If you write a common code that should get compiled or different platforms, add `org.jetbrains.kotlinx:atomicfu-common`
+If you write a common code that should get compiled or different platforms, add `org.jetbrains.kotlinx:atomicfu`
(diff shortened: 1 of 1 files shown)


biberesser:
This hast cost me half an hour of my life :(
Thanks for the solution/workaround!

forrestpangborn:
> This hast cost me half an hour of my life :(
> Thanks for the solution/workaround!

It wasted about 3 hours of mine due to trying to sort it out on a very large project with long build times 😆 Figured I could at least try to fix it to help others avoid the same situation!

mvicsokolova (approved):

Issue #122, opened 2020-02-17
Support Android Native targets


elizarov:
This is needed for https://github.com/Kotlin/kotlinx.coroutines/issues/812
See also https://github.com/Kotlin/kotlinx.coroutines/issues/1808

elizarov:
This shoud be as easy as adding the following lines to [`atomicfu/build.gradle`](https://github.com/Kotlin/kotlinx.atomicfu/blob/master/atomicfu/build.gradle):
```
            addTarget(presets.androidNativeArm32)
            addTarget(presets.androidNativeArm64)
            addTarget(presets.androidNativeX86)
            addTarget(presets.androidNativeX64)
```
It builds successfully. But we need to test it somehow.

piiertho:
Hi @elizarov !

I guess I will try to compile it and use it on my kotlin wrapper for Godot in a first time.
Then I'll see if I have an idea to make test and run them in CI.

You talked about infrastructure limitation to do the test, can you give some insight about those limitations ?

elizarov:
We have not explored yet how to setup a test for the Android/Native code that is compiled that way. Propbably, it needed to be run under Android emulateor and we have not looked yet at how we can make it happen on our CI via Grable.

piiertho:
> We have not explored yet how to setup a test for the Android/Native code that is compiled that way. Propbably, it needed to be run under Android emulateor and we have not looked yet at how we can make it happen on our CI via Grable.

That was my first thinking. I'll try it as soon as I d'une the time to do it :)

natario1:
Is there any progress on this issue? Are you using your own internal CI? How can one contribute exactly?

elizarov:
We don't have resources to work on it at the moment.

glootjofo:
I think a lot of people want to help on this, but seeing your CI pipeline is closed, could you perhaps either make it possible for people to contribute to it, or just swap to github actions? There is a lot of will, but we can't help you guys, and it sucks.

mvicsokolova:
We are working internally on guarantees that we are ready to provide for each individual target (as well as the overall understanding which targets we are going to support) and as soon as the public guarantees are published, we'll add each supported target to all our libraries at once

PR #206, opened 2021-10-21
Add androidNative targets + test plugin

natario1:
See discussion: https://github.com/Kotlin/kotlinx.atomicfu/issues/122#issuecomment-948865934

To run tests,
- set either $ANDROID_HOME or `multiplatformTesting.androidTools.sdkHome` to the Android SDK directory or, if not installed, to some directory where the plugin can download them
- run `./gradlew atomicfu:runAllAndroidNativeTests`

Testing plugin readme: https://github.com/deepmedia/multiplatform-testing/blob/main/readme.md

### DIFF
--- a/atomicfu/build.gradle
+++ b/atomicfu/build.gradle
@@ -3,6 +3,7 @@
  */
 
 apply plugin: 'kotlin-multiplatform'
+apply plugin: 'io.deepmedia.tools.multiplatform-testing' // androidNative test tasks
 apply from: rootProject.file("gradle/targets.gradle")
 
 ext {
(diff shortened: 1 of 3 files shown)


natario1:
Hi @mvicsokolova , while still hoping that this gets merged soon (I am happy to provide any support)...

Is it OK for you if I publish these artifacts myself in Maven Central, with slightly different coordinates (like `my.company.org.jetbrains.kotlinx:atomicfu` instead of `org.jetbrains.kotlinx:atomicfu`)?

The difference would be that my publication would have Android Native targets, while yours hasn't. Then I can do the same for coroutines and write a simple gradle plugin for dependency substitution for end users. From my understanding, the license allows this, but it would be nice to have your approval, even if informally.

roy-shchory (changes requested):
What about arm 32 😍

natario1:
Arm32 can be added after Kotlin 1.6.20, which contains an important fix for 32bit atomics.

mvicsokolova:
We postpone the support of android native targets till the guarantees are published for them.
Please see [#122 comment](https://github.com/Kotlin/kotlinx-atomicfu/issues/122#issuecomment-1300306280)

PR #215, opened 2022-01-25
Atomicfu Js/Ir compiler plugin supported in the gradle plugin

mvicsokolova:
- Added support of the compiler plugin transformation to the `AtomicfuGradlePlugin`.
- `atomicfu-gradle-plugin` test projects are ignored for a while. I'm going to create a PR with new test projects soon.
- Kotlin version is for now `1.7.0-dev-565`, [built](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_DeployMavenArtifacts/160253827?showRootCauses=false&expandBuildChangesSection=true) from the `1.6.20-M1-release branch`

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ dependencies {
 
 ## IR transformation for Kotlin/JS
 
-There is a new option to turn on IR transformation for Kotlin/Js backend.
+There is a new option to turn on IR transformation for Kotlin/JS backend.
 You can add `kotlinx.atomicfu.enableIrTransformation=true` to your `gradle.properties` file in order to enable it.
 
 Here is how transformation is performed for different [JS compiler modes](https://kotlinlang.org/docs/js-ir-compiler.html) with this option enabled:
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ apply plugin: 'kotlinx-atomicfu'
 
 ### JS 
 
-Configure add apply plugin just like for [JVM](#jvm). 
+Configure add apply plugin just like for [JVM](#jvm).
 
 ### Native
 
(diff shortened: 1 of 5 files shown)


olme04:
@mvicsokolova just for curiosity, what is the main difference between JS and IR variants?

mvicsokolova:
Hi, @olme04!
In case of `JS` option the transformation is applied after the compilation to the generated *.js files and JavaScript transformer is quite fragile as it depends on the JS syntax.

Whereas, for `IR` option the transformation is applied to the IR and relies to the compiler API of the IR tree, that is a more stable approach. So, IR option  is preferable if the Kotlin/JS IR backend is used for compilation.

mvicsokolova:
Atomicfu compiler plugin for Kotlin/JS IR backend will be supported since Kotlin `1.6.20`

olme04:
@mvicsokolova, thx for the answer!
One more question - what should be the best variant, when using `js(BOTH)`? (my used case is library, to run tests on both legacy and IR)

ilgonmic (approved):

mvicsokolova:
I’ve figured, that we can only obtain values passed to the `atomic {}` extension after project evaluation, though at the same time we need those values (particularly `jsVariant` = `IR`/`JS` ) during task configuration, that should be done earlier. 

Maybe there is some other way to set a configuration parameter before the project is evaluated?

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -33,10 +33,7 @@ open class AtomicFUGradlePlugin : Plugin<Project> {
         val pluginVersion = rootProject.buildscript.configurations.findByName("classpath")
             ?.allDependencies?.find { it.name == "atomicfu-gradle-plugin" }?.version
         extensions.add(EXTENSION_NAME, AtomicFUPluginExtension(pluginVersion))
-        if (config.transformJs && config.jsVariant == "IR") {
-            // apply the compiler plugin, if IR transformation should be applied to KotlinJsIrTarget compilations
-            plugins.apply(AtomicfuKotlinGradleSubplugin::class.java)
-        }
+        plugins.apply(AtomicfuKotlinGradleSubplugin::class.java)
```
mvicsokolova, line 36: I’ve figured, that we can only obtain values passed to the `atomic {}` extension after project evaluation, though `jsVariant` option was supposed to define whether to apply the compiler plugin (`AtomicfuKotlinGradleSubplugin`) and compiler plugin should be applied before project evaluation is completed. 

For now, compiler plugin is applied by default.

Maybe there is some other way to create a user flag, that can be resolved before the project is evaluated?

README.md
```diff
@@ -149,16 +149,49 @@ dependencies {
 }
 ```
 
-### Additional configuration
+## IR transformation for Kotlin/JS
 
-There are the following additional parameters (with their defaults):
+There is a new option to turn on IR transformation for Kotlin/Js backend.
+You should add `kotlinx.atomicfu.enableIrTransformation=true` to your `gradle.properties` file.
```
qwwdfsad, line 155:
```suggestion
You can add `kotlinx.atomicfu.enableIrTransformation=true` to your `gradle.properties` file in order to enable it.
```

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -18,18 +18,25 @@ import org.jetbrains.kotlin.gradle.plugin.*
...
 private const val IMPLEMENTATION_CONFIGURATION = "implementation"
 private const val TEST_IMPLEMENTATION_CONFIGURATION = "testImplementation"
+private const val ENABLE_IR_TRANSFORMATION = "kotlinx.atomicfu.enableIrTransformation"
 
 open class AtomicFUGradlePlugin : Plugin<Project> {
     override fun apply(project: Project) = project.run {
         val pluginVersion = rootProject.buildscript.configurations.findByName("classpath")
             ?.allDependencies?.find { it.name == "atomicfu-gradle-plugin" }?.version
         extensions.add(EXTENSION_NAME, AtomicFUPluginExtension(pluginVersion))
+        if (rootProject.findProperty(ENABLE_IR_TRANSFORMATION).toString().toBoolean()) {
```
qwwdfsad, line 37: You probably need `toBooleanStrict` here to prevent typos coercing into `false`

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -129,9 +158,9 @@ fun Project.withKotlinTargets(fn: (KotlinTarget) -> Unit) {
     extensions.findByType(KotlinProjectExtension::class.java)?.let { kotlinExtension ->
         val targetsExtension = (kotlinExtension as? ExtensionAware)?.extensions?.findByName("targets")
         @Suppress("UNCHECKED_CAST")
-        val targets = targetsExtension as NamedDomainObjectContainer<KotlinTarget>
+        val targets = targetsExtension as? NamedDomainObjectContainer<KotlinTarget>
```
qwwdfsad, line 161: Could you please elaborate why null-safety is required here?

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
mvicsokolova: Gradle seems to force stdlib version to 1.4, the same as bundled in it's runtime. So using `toBooleanStrict() `fails with unresolved reference. And forcing stdlib version in `atomicfu-gradle-plugin` does not help. Maybe we should update gradle version?)

qwwdfsad: I believe it's safer to copy-paster `toBooleanStrict` implementation here

PR #216, opened 2022-01-26
Сonfiguration cache bug caused by capturing the compilation object into lambda.

### LINKED ISSUE #212: Configuration cache problem when applying kotlinx-atomicfu Gradle plugin along with kotlin-multiplatform
Original issue: https://youtrack.jetbrains.com/issue/KT-49165

There's a [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) problem in kotlin-multiplatform test compilation task configuration: https://github.com/Kotlin/kotlinx.atomicfu/blob/1c9f04c3d584e2e6d8ceab133abb8b19d3c0ff51/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt#L217
`compilation` object shouldn't be referenced from task action (the lambda passed into `doFirst`). It causes `compilation` object to be captured into the lambda and Gradle tries to serialize it.

Reproducer project: [KT-49165.zip](https://github.com/Kotlin/kotlinx.atomicfu/files/7596349/KT-49165.zip)

Configuration cache report: [configuration-cache-report.html.zip](https://github.com/Kotlin/kotlinx.atomicfu/files/7596359/configuration-cache-report.html.zip)

mvicsokolova:
Fixes #212

### DIFF
--- a/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
+++ b/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
@@ -211,8 +211,9 @@ private fun Project.configureTransformationForTarget(target: KotlinTarget) {
             (tasks.findByName("${target.name}${compilation.name.capitalize()}") as? Test)?.classpath =
                 originalMainClassesDirs + (compilation as KotlinCompilationToRunnableFiles).runtimeDependencyFiles - mainCompilation.output.classesDirs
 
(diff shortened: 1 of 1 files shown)


qwwdfsad:
One of the tests seems to fail now

mvicsokolova:
Yes, `MppProjectTest` fails because task outputs are not cached when `build` is called for the second time 😕
I'm trying to figure out what's wrong.

ALikhachev:
Isn't it possible to add friend paths at configuration time?
Something like 
```kotlin
compilation.compileKotlinTaskProvider.configure {
    it.kotlinOptions.addFriendPaths(originalMainClassesDirs)
}
```

PR #217, opened 2022-02-01
Updated tests for atomicfu-gradle-plugin

mvicsokolova:
The base branch is `js-plugin-fix` because `JsProjectTest` needs the fix in AtomicfuGradlePlugin for Kotlin/JS projects (#218)


ilya-g (approved):
Minor: some empty files were left, probably they are not needed anymore.

PR #242, opened 2022-08-22
NonReentrantLock with conditions and switch to new memory model

qwwdfsad:


qwwdfsad:
No need to hurry with the review tho.
I've prototyped `BlockingQueue` on top of that and intend it to merge no sooner than my CR in coroutines is ready

qwwdfsad:
It seems not to be the way, let's postpone it for a bit

qwwdfsad:
This change is not needed at all for coroutines

PR #244, opened 2022-09-14
Update gradle version to 6.8.3

mvicsokolova:


qwwdfsad (approved):
Please elaborate on the reasons

mvicsokolova:
In Kotlin master the minimal gradle version was updated to `6.8.3`.
So, now when building against master, we are getting the error:
```
GradleException: The current Gradle version 6.8.1 is not compatible with the Kotlin Gradle plugin. Please use Gradle 6.8.3 or newer, or the previous version of the Kotlin plugin.
```

And then we will get this error for `1.8.0-Beta`, `1.8.0`

qwwdfsad:
👍 works for me

PR #265, opened 2022-11-25
Prepare atomicfu build scripts for inclusion into the main aggregate as a user project

anastasiiaSpaseeva:


atomicfu-gradle-plugin/build.gradle
```diff
@@ -14,8 +14,8 @@ if (rootProject.ext.jvm_ir_enabled) {
 // Gradle plugin must be compiled targeting the same Kotlin version as used by Gradle
 kotlin.sourceSets.all {
     languageSettings {
-        apiVersion = "1.4"
-        languageVersion = "1.4"
+        apiVersion = rootProject.ext.kotlin_api_version
```
qwwdfsad, line 17: This is a bit too unmaintainable for us -- both in terms of readability (i.e. _why_ the code is written this way) and writability (i.e. where to update the language version).

I suggest making rr build explicit:
```
languageSettings {
    apiVersion = getApiVersionForOverriddenKotlin() ?: "1.4"
    ... same pattern everywhere else...
}
```

This way we know **why** the version can be overridden, while still having the actual default where it is supposed to be

atomicfu-maven-plugin/build.gradle
```diff
@@ -23,6 +23,10 @@ def buildSnapshots = rootProject.properties['build_snapshot_train'] != null
 
 evaluationDependsOn(':atomicfu-transformer')
 
+def kotlinDevRepoUrl = (rootProject.properties['kotlin_repo_url'] ?: "" ) as String
+println("Configured external repository url: '$kotlinDevRepoUrl'")
```
qwwdfsad, line 27: Please avoid `println` calls enabled by default

atomicfu-maven-plugin/build.gradle
```diff
@@ -23,6 +23,10 @@ def buildSnapshots = rootProject.properties['build_snapshot_train'] != null
 
 evaluationDependsOn(':atomicfu-transformer')
 
+def kotlinDevRepoUrl = (rootProject.properties['kotlin_repo_url'] ?: "" ) as String
```
qwwdfsad, line 26: We should have a comment what this is for

build.gradle
```diff
@@ -28,13 +28,15 @@ buildscript {
     // These two flags are enabled in train builds for JVM IR compiler testing
     ext.jvm_ir_enabled = rootProject.properties['enable_jvm_ir'] != null
     ext.native_targets_enabled = rootProject.properties['disable_native_targets'] == null
+    def kotlinDevRepoUrl = (rootProject.properties['kotlin_repo_url'] ?: "" ) as String
```
qwwdfsad, line 31: Same here

qwwdfsad (changes requested):
* Please change the target branch to `develop`
* Currently, all RR/train support is stretched through all Gradle files. It's both hard to maintain and understand. Please create a separate `.kts` file and declare all RR-specific variables and functions there. Then on call-sites we will have clear entry-point:

```
// KotlinDevSupport.kts
// Here will be a huge comment explaining why this is needed and how it is used

val kotlinDevRepo = ...
fun getApiVersionForOverriddenKotlin() = ...
```

Pros: everything is documented in a single place, easy to maintain, hard to miss, no questions on individual call sites

atomicfu-maven-plugin/build.gradle
anastasiiaSpaseeva: I've added a kdoc in KotlinAggregateBuild.kt

qwwdfsad (approved):

PR #284, opened 2023-03-17
Removed gradle plugin application tests for JS legacy backend

mvicsokolova:


mvicsokolova:
Actually, there was a wrong motivation for these changes, the build of master [failed](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxAtomicfu_BuildAggregatedLinuxWindows/4099945?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&expandBuildTestsSection=true) with an error `Could not find org.nodejs:node:16.13.0` -- nothing to do with js legacy backend. So, It's not necessary to merge it now, we can do it when legacy JS backend will be deprecated in Kotlin 1.9.

Issue #295, opened 2023-04-20
JVM legacy transformation: both original and transformed classes stay in the test classpath


mvicsokolova:
The issue is reproduced with `JvmLegacyTransformationTest.testClasspath` for Kotlin `1.8.20+`.
Both original and transformed classes stay in the test compile/runtime classpath, though just transformed classes should be there.

This is the issue only for JVM projects that use legacy bytecode transformation.

mvicsokolova:
Fixed in this PR: #303

PR #296, opened 2023-04-25
Remove JS Legacy configurations (updated)

mvicsokolova:


atomicfu/build.gradle
```diff
@@ -17,6 +19,12 @@ ext {
     }
 }
 
+// TODO: this block should be removed after the switch to Kotlin language version 1.9
+// (right now it implements a toggle between IR and BOTH JS compiler types for testing of Kotlin dev builds)
+def fromVersion = { String arg -> KotlinVersion.values().find { it.version == arg } }
+def kotlinLanguageVersion = fromVersion(KotlinAggregateBuild.getOverriddenKotlinLanguageVersion(project) ?: "1.4")
+def isJsLegacyCompilerEnabled = kotlinLanguageVersion < KotlinVersion.KOTLIN_1_9
```
Tapchicoma, line 26: LanguageVersion has nothing with JS target deprecation. Should be checked using KGP version.

atomicfu/build.gradle
```diff
@@ -153,63 +164,68 @@ dependencies {
...
+    def compileJsLegacy = tasks.hasProperty("compileKotlinJsLegacy")
+            ? compileKotlinJsLegacy
+            : compileKotlinJs
 
-// Workaround the problem with Node downloading
-repositories.whenObjectAdded {
-    if (it instanceof IvyArtifactRepository) {
-        metadataSources {
-            artifact()
+    tasks.withType(compileJsLegacy.getClass()) {
```
Tapchicoma, line 177: - `compileJsLegacy` should be already a `Class`.
- after `.withType` `.configureEach { ... }` should be used

atomicfu/build.gradle
FenstonSingel: I made these changes based on the fact that CLI-level deprecation of JS/Legacy [seems to work based on the language version at the moment](https://github.com/JetBrains/kotlin/blob/552e640517ffa10ab60ba4c6af61e5f18d71cc35/compiler/cli/cli-js/src/org/jetbrains/kotlin/cli/js/K2JSCompiler.java#L189), so we might want to take a closer look at it in the compiler as well then.

mvicsokolova: Hmm 👀 
For now I rolled this back to KGP version check

atomicfu/build.gradle
mvicsokolova: Fixed

atomicfu/build.gradle
```diff
@@ -17,6 +19,10 @@ ext {
     }
 }
 
+// TODO: this block should be removed after the switch to Kotlin language version 1.9
```
FenstonSingel, line 22: If we settle on KGP version check in the end, corresponding TODOs (three of them in `gradle.build`) should probably also be updated (changed from "Kotlin language version 1.9" to "Kotlin release version 1.9.0" or smth) to escape potential future confusion when migrating to KGP 1.9.0 even without changing the language version.

FenstonSingel:
I've tried compiling this branch with Kotlin 1.9.0-dev-6688 and various LVs. The results are:

- for LV 1.4 (project LV), only one test fails (it is tracked by #295 and is muted in relevant TC configurations)
- for LV 1.9, the results are the same as for LV 1.4
- for LV 2.0, tests for post-compilation JVM transformers (`transformedTestXX`) fail (it should be fixed when the next kotlinx.metadata release is published and migrated to) and 11 out of 11 completed `atomicfu-gradle-plugin` tests fail (they are muted in relevant TC configurations so I haven't taken a closer look)

From testing of Kotlin dev-builds's perspective, everything is great for 1.9.0-dev TC configurations and good enough for K2 TC configurations (we will probably mute all tests in relevant TC configurations eventually regardless, so red tests are not a big problem).

mvicsokolova:
Thank you! 

I'll have a look what is wrong with `atomicfu-gradle-plugin` tests for lv=2.0

mvicsokolova:
I can locally reproduce the failure of `atomicfu-gradle-plugin` tests with lv = 1.9. They fail with `Could not initialize class kotlinx.atomicfu.plugin.gradle.Platform`. Probably that is somehow related to EnumEntries (e.g. [KT-57317](https://youtrack.jetbrains.com/issue/KT-57317)).

buildSrc/src/main/kotlin/KotlinVersion.kt
```diff
@@ -0,0 +1,11 @@
+@file:JvmName("KotlinVersion")
+
+import org.gradle.api.*
+
+fun isKotlinVersionAtLeast(kotlinVersion: String, atLeastMajor: Int, atLeastMinor: Int): Boolean {
+    val (major, minor) = kotlinVersion
+        .split('.')
+        .take(2)
+        .map { it.toInt() }
+    return major == atLeastMajor && minor >= atLeastMinor
```
FenstonSingel, line 10: If conditional removal of JS/Legacy persists up until migration to 2.0.0 (or 2.0.0-dev), JS/Legacy will be enabled back after the switch to 2.0.0 as `isKotlinVersionAtLeast` considers `2.0 < 1.9` right now due to equality comparison on major version. I suggest the following tweak for this line:
```kotlin
return when {
    major > atLeastMajor -> true
    major < atLeastMajor -> false
    else -> minor >= atLeastMinor
}
```

buildSrc/src/main/kotlin/KotlinVersion.kt
```diff
@@ -0,0 +1,11 @@
+@file:JvmName("KotlinVersion")
+
+import org.gradle.api.*
```
FenstonSingel, line 3: Unused import.

FenstonSingel:
> @FenstonSingel Am I right, that we can merge these change into develop now? And for dev-builds you will be able to mute failing tests for now?

Yes, any failing tests are not a blocker for dev-builds testing.

buildSrc/src/main/kotlin/KotlinVersion.kt
mvicsokolova: Fixed

FenstonSingel (approved):

atomicfu/build.gradle
mvicsokolova: With gradle 6.8.3 compilation fails if I remove getClass()

Tapchicoma: I mean in the lines above it could be:
```groovy
    def compileJsLegacy = tasks.hasProperty("compileKotlinJsLegacy")
            ? compileKotlinJsLegacy.getClass()
            : compileKotlinJs.getClass()
```

PR #297, opened 2023-04-25
Fix build error on using KGP removed method

Tapchicoma:
Related YT issue: https://youtrack.jetbrains.com/issue/KT-53748/Remove-KotlinCompile-setClasspath-getClasspath-methods


mvicsokolova (approved):

mvicsokolova:
I think, we should put this commit to `develop` and then rebase `kotlin-community/dev` on it

Tapchicoma:
I need it for my MR into Kotlin. I will create a separate PR into `develop` cherry-picking this commit

PR #303, opened 2023-05-08
Duplicated class files in jar

mvicsokolova:
After update from Gradle 6.8 to 7.3 the problem with building jvm-only projects was found: jar task is failing with this error
```
Execution failed for task ':jar'.
> Entry IntArithmetic.class is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.3/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.
```

Reproduced with Kotlin 1.8.20+

Jar task tries to pack both original and transformed class files.

Fixes #301, #295

PR #305, opened 2023-05-25
Version 0.21.0

mvicsokolova:


qwwdfsad:
Some observations about the commit history we got here: there is a sequence of commits that exist out-of-context (e.g. neither tested nor thoroughly explained (e.g. 5f05039ed7d929942619158e2fb7288412f66dae), with the final issue that is fixed also non-tested. They should've been squashed, with the detailed explanation why the changes are done.

I'd suggest using https://github.com/Kotlin/kotlinx-kover/pull/386 as an inspiration: the Gradle tweaks are documented, the test is provided, the commit message is extensive (e.g. *why* file collection was replaced with the provider, not the fact it was done)

mvicsokolova:
Thank you, yes, all these out-of-context commits from [c201c53](https://github.com/Kotlin/kotlinx-atomicfu/pull/305/commits/c201c5386ff749fdcc81250e257614a84b4a54ce) to [c41d50a](https://github.com/Kotlin/kotlinx-atomicfu/pull/305/commits/c41d50afeaf65465974cb3f028c19cf7810d3aa4) were merged with this PR: https://github.com/Kotlin/kotlinx-atomicfu/pull/303. My mistake they were not squashed. The issue fixed there was reproduced by the existing test `kotlinx.atomicfu.plugin.gradle.test.JvmLegacyTransformationTest`.

I can fix the history in this branch and merge it to master in a proper way (squash + extended comment).

I would also add comments to all the changes for future understanding (some of them were just caused by using the new KGP API).

mvicsokolova:
Currently, the release is blocked by this issue #307 that is fixed here #308 (waiting for review and approval).
It is only reproduced for now with Kotlin `1.9.0-RC`, but I want to add the fix to this release.

mvicsokolova:
Commit fixing the issue with building JVM-only projects merged ([88a5e14](https://github.com/Kotlin/kotlinx-atomicfu/pull/305/commits/88a5e144c948f46697d80d26dcb0a2a2d5b9b8e7)). 

`develop` history fixed

Issue #313, opened 2023-06-18
Update native atomics implementation after update to Kotlin 1.9.0


mvicsokolova:
Migrate from deprecated native `kotlin.native.concurrent` atomics to new `kotlin.concurrent` atomics.

PR #314, opened 2023-06-18
Update native atomics implementation

mvicsokolova:
This commit replaces deprecated native atomics from `kotlin.native.concurrent` package with new atomics from `kotlin.concurrent`, because deprecation levels of native atomics will be updated for 1.9.20 in this MR (https://jetbrains.team/p/kt/reviews/10650/timeline), see KT-58123. These changes are required in `kotlin-community/dev` branch for the aggregate build to pass.

NOTE: this commit is not present in develop branch, because new `kotlin.concurrent` atomics are only available since Kotlin 1.9.0. IT SHOULD BE REMOVED after kotlinx.atomicfu updates Kotlin version to 1.9.0 (#313)

### DIFF
--- a/atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
+++ b/atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
@@ -6,9 +6,9 @@
 
 package kotlinx.atomicfu
 
-import kotlin.native.concurrent.AtomicInt as KAtomicInt
-import kotlin.native.concurrent.AtomicLong as KAtomicLong
-import kotlin.native.concurrent.FreezableAtomicReference as KAtomicRef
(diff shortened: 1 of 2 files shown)


mvicsokolova:
The commit from this PR was reverted, because it caused the aggregate build to fail (apiVersion = 1.9 was required to use new native atomics, but it was 1.4 in the project), thus I've now reopened the PR and commited the apiVersion update.

This is the [test aggregate build](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_kotlinxAtomicfu/322679957?hideProblemsFromDependencies=false&hideTestsFromDependencies=false) to ensure that changes from the current kotlinx.atomicfu branch are enough for the MR (https://jetbrains.team/p/kt/reviews/10650/timeline)

This problem could be revealed before merge if `kotlin-community/dev` branch relied on the existing Kotlin version, rather then dev versions that may be already removed, I've filed the issue for that: #315.

PR #319, opened 2023-06-26
Update gradle version to 8.1

mvicsokolova:
Updated gradle to 8.1

* Removed deprecated experimentalWarning from kotlinDslOptions 
* I had to manually set `jvmToolChain(8)` for JVM target compatibility in all subprojects, because for some reason this [config](https://github.com/Kotlin/kotlinx-atomicfu/blob/f8a2a75685bd71ac290ce1bc869ed98f46895616/build.gradle#L76) was not enough and the build was failing with errors like this:
```
Execution failed for task ':atomicfu-gradle-plugin:compileKotlin'.
> 'compileJava' task (current target is 1.8) and 'compileKotlin' task (current target is 11) jvm target compatibility should be set to the same Java version.
  Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

### DIFF
--- a/atomicfu-gradle-plugin/build.gradle
+++ b/atomicfu-gradle-plugin/build.gradle
@@ -98,3 +98,7 @@ if (isKotlinVersionAtLeast(ext.kotlin_version, 1, 9)) {
         exclude "**/JsLegacyTransformationTest*", "**/MppLegacyTransformationTest*"
     }
 }
+
+kotlin {
+    jvmToolchain(8)
+}
--- a/atomicfu-gradle-plugin/build.gradle
+++ b/atomicfu-gradle-plugin/build.gradle
@@ -98,7 +98,3 @@ if (isKotlinVersionAtLeast(ext.kotlin_version, 1, 9)) {
         exclude "**/JsLegacyTransformationTest*", "**/MppLegacyTransformationTest*"
     }
 }
-
-kotlin {
-    jvmToolchain(8)
-}
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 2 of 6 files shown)


atomicfu-gradle-plugin/build.gradle
```diff
@@ -98,3 +98,7 @@ if (isKotlinVersionAtLeast(ext.kotlin_version, 1, 9)) {
         exclude "**/JsLegacyTransformationTest*", "**/MppLegacyTransformationTest*"
     }
 }
+
+kotlin {
```
fzhinkin, line 102: Replacing 
```
sourceCompatibility = JavaVersion.VERSION_1_8
targetCompatibility = JavaVersion.VERSION_1_8
```

with 
```
toolchain {
    languageVersion.set(JavaLanguageVersion.of(8))
}
```
in the root build.gradle file should solve the issue without specifying `jvmToolchain` for each individual plugin.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.1-all.zip
```
fzhinkin, line 5: What's the reason for updating the version?

mvicsokolova: The reason is preparation of kotlinx.atomicfu for K2. Expected Gradle version to be compatible with K2 is 8.3, so we decided to steadily update Gradle version first on `develop` and then try out the 8.3-nightly version in the `kotlin-community/k2/dev` branch.

atomicfu-gradle-plugin/build.gradle
mvicsokolova: fixed

gradle/wrapper/gradle-wrapper.properties
fzhinkin: Could you please mention that in the commit message when merging the PR? Thanks!

PR #325, opened 2023-07-24
Comply with new compiler restriction on actual declaration annotations

merfemor:
In [KT-58551](https://youtrack.jetbrains.com/issue/KT-58551) we require all annotations from expect declaration to be present on actual.
This commit fixes violations on get/setValue methods, having `@InlineOnly` set only on `expect` declaration.


mvicsokolova (approved):

mvicsokolova:
Should it also appear in the kotlin-community/dev branch for the aggregate build?

PR #334, opened 2023-08-14
Add WebAssembly target

igoriakovlev:
Same implementation as in K/JS but without @JsName annotations


atomicfu/build.gradle
```diff
@@ -49,6 +49,10 @@ kotlin {
     // JVM -- always
     jvm()
 
+    wasm() {
+      d8()
```
fzhinkin, line 53: Why it is `d8` and not `nodejs`, like in the JS case?

mvicsokolova:
Could you please also elaborate on the publication process of `wasm` artifacts?

For now, Deploy build has always been [red](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxAtomicfu_DeployWasmLinuxAll/4275364?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildProblemsSection=true), as wasm target was published from the separate branch. Is the configuration going to be fixed?

atomicfu/src/wasmMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
```diff
@@ -0,0 +1,178 @@
+/*
```
fzhinkin, line 1: Is it possible to simply set the Wasm's sources dir to the  JS source's dir instead of copying files? Or there's a downside of compiling code w/ `JsName` into Wasm?

mvicsokolova:
Can we also set up a test build of coroutines with the enabled wasm target and atomicfu dependency before merging this, so that we are sure that no transformations are applied, all dependencies are present and all the necessary artefacts are published?

atomicfu/build.gradle
igoriakovlev: It does not matter really which engine we use in this case, d8 and nodejs both use v8 under the hood.

atomicfu/src/wasmMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
igoriakovlev: We could try to do that using intermediate MPP sourceset (but then we need to disable metadata tasks) or include shared directory onto two sourcets. In both cases JsName have to be typealiased in JS and stubbed in Wasm (expect/actual or not). There was not so much copy-pasted code so I decided for a while not to do that. Do you think it worth?

igoriakovlev:
> Can we also set up a test build of coroutines with the enabled wasm target and atomicfu dependency before merging this, so that we are sure that no transformations are applied, all dependencies are present and all the necessary artefacts are published?

I do not know how to do that. I have already made [MR](https://github.com/Kotlin/kotlinx.coroutines/pull/3849) for coroutines and this coroutines branch works perfectly with this Wasm AtomicFU on local machine (using train build). But I dont know how to use this train mode on CI.

atomicfu/src/wasmMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
fzhinkin: It's worth reducing the amount of duplicated code.

I tried to simply set Wasm's sourceDir to Js's one and everything worked fine without:
> stubbed in Wasm (expect/actual or without).

Or you meant that it's already done?

atomicfu/build.gradle
fzhinkin: Why not use the same engine then? It looks a bit confusing

igoriakovlev: In wasm we are trying to use d8 because nodejs's v8 is always little outdated. But for here for kotlin 1.9 we can use nodejs as well. I can replace it with nodejs.

atomicfu/src/wasmMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
igoriakovlev: Its better NOT to mark Wasm declarations with JsName in this case. So for Wasm target this annotation should be stubbed. I can do it and you will see what I mean.

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/OptionalJsName.kt
```diff
@@ -0,0 +1,14 @@
...
+@Retention(AnnotationRetention.BINARY)
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.CONSTRUCTOR,
+    AnnotationTarget.PROPERTY_GETTER,
+    AnnotationTarget.PROPERTY_SETTER
+)
+expect annotation class OptionalJsName(val name: String)
```
fzhinkin, line 14: It would be nice to have a comment clarifying purpose of this annotation (i.e. it's aliased by the `JsName` when compiled to JS and it's NOOP when compiled to Wasm).
Otherwise it's unclear why the name is optional.

igoriakovlev: Done

mvicsokolova:
Could you please add a test project for `atomicfu-gradle-plugin`, simillar to this JS project test: https://github.com/Kotlin/kotlinx-atomicfu/blob/3a3deb61243d69740fbded19f02472873e6942b9/atomicfu-gradle-plugin/src/test/kotlin/kotlinx/atomicfu/plugin/gradle/test/JsProjectTest.kt

- to be sure that when the plugin is applied, no transformations are performed, checking that only compile tasks are [present](https://github.com/Kotlin/kotlinx-atomicfu/blob/3a3deb61243d69740fbded19f02472873e6942b9/atomicfu-gradle-plugin/src/test/kotlin/kotlinx/atomicfu/plugin/gradle/test/JsProjectTest.kt#L36C3-L36C3)
- to check that atomicfu-wasm dependency is present in both compile and runtime classpath

In overall all the changes look good to me.

hfhbd:
Do you also have plans to add wasmWasi?

fzhinkin:
BTW, should we also update the README to mention the new target there?
@mvicsokolova

mvicsokolova:
Rebased the branch on the latest develop branch with Integration tests

mvicsokolova:
@hfhbd yes, I've added support of `wasmWasi` target as well 🙏🏻

mvicsokolova:
I've introduced `wasmJs` and `wasmWasi` targets in integration tests: they check atomicfu-gradle-plugin application to the mpp project with both wasm targets declared.

For wasm targets `atomicfu-gradle-plugin` does not perform any transformations, just adds the implementation kotlinx-atomicfu dependency.

mvicsokolova:
Mentioned in README that the library provides support of all platforms including Wasm (since Kotlin 1.9.20)

mvicsokolova (approved):

build.gradle
```diff
@@ -107,3 +105,15 @@ if (build_snapshot_train) {
...
+extensions.findByType(org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension.class).with {
+    // canary nodejs that supports recent Wasm GC changes
+    it.nodeVersion = '21.0.0-v8-canary202309167e82ab1fa2'
+    it.nodeDownloadBaseUrl = 'https://nodejs.org/download/v8-canary'
+}
+
+// We need to ignore unsupported engines (i.e. canary) for npm
+tasks.withType(org.jetbrains.kotlin.gradle.targets.js.npm.tasks.KotlinNpmInstallTask).configureEach {
+    args.add("--ignore-engines")
+}
```
fzhinkin, line 119:
```suggestion
}

```

atomicfu/src/jsMain/kotlin/kotlinx/atomicfu/OptionalJsName.kt
```diff
@@ -0,0 +1,3 @@
+package kotlinx.atomicfu
+
+actual typealias OptionalJsName = JsName
```
fzhinkin, line 3:
```suggestion
actual typealias OptionalJsName = JsName

```

README.md
```diff
@@ -43,11 +43,9 @@ Starting from version `0.22.0` of the library your project is required to use:
 ## Features
 
 * Code it like a boxed value `atomic(0)`, but run it in production efficiently:
-  * as `java.util.concurrent.atomic.AtomicXxxFieldUpdater` on Kotlin/JVM 
+  * as `java.util.concurrent.atomic.AtomicXxxFieldUpdater` on Kotlin/JVM
   * as a plain unboxed value on Kotlin/JS
-* Multiplatform: write common Kotlin code with atomics that compiles for Kotlin JVM, JS, and Native backends:
-    * Compile-only dependency for JVM and JS (no runtime dependencies)
-    * Compile and runtime dependency for Kotlin/Native 
+* Complete multiplatform support: JVM, Native, JS and Wasm (since Kotlin 1.9.20).
```
fzhinkin, line 48: @mvicsokolova does it make sense to clarify how different transformation flags  (from the [Transformation modes](https://github.com/Kotlin/kotlinx-atomicfu/blob/e10743be50e1caeb592679fab0618061820a115c/README.md#transformation-modes) and the [Options for post-compilation transformation](https://github.com/Kotlin/kotlinx-atomicfu/pull/334/files#options-for-post-compilation-transformation)) should work with Wasm?

mvicsokolova: Yes, I think it's good to drop a word that the plugin does not affect Wasm target at all, and no transformations are applied.

atomicfu/src/jsAndWasmSharedMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -2,9 +2,10 @@ package kotlinx.atomicfu.locks
 
 import kotlinx.atomicfu.REENTRANT_LOCK
 
+@Suppress("ACTUAL_CLASSIFIER_MUST_HAVE_THE_SAME_MEMBERS_AS_NON_FINAL_EXPECT_CLASSIFIER")
```
fzhinkin, line 5:
```suggestion
@Suppress("ACTUAL_CLASSIFIER_MUST_HAVE_THE_SAME_MEMBERS_AS_NON_FINAL_EXPECT_CLASSIFIER_WARNING")
```

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
```diff
@@ -41,4 +41,11 @@ class MppProjectTest {
         assert(mppSample.cleanAndBuild().isSuccessful)
         mppSample.checkConsumableDependencies()
     }
+    
+    @Test
+    fun testMppWasmBuild() {
+        assert(mppSample.cleanAndBuild().isSuccessful)
```
fzhinkin, line 47: I missed it during the integration tests review: please use `assertTrue` here and in other tests instead of `assert`. Behavior of the `assert` depends on runtime flags and for testing purposes the `kotlinx.test.assertTrue` is much better suited.

atomicfu/src/jsAndWasmSharedMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
mvicsokolova: But `kotlin.AllowDifferentMembersInActual` is not resolved for some reason, though dependency to `kotlin-stdlib-wasm-js` is provided.

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
mvicsokolova: Fixed

atomicfu/src/jsAndWasmSharedMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
igoriakovlev: Just remove this suppress. `AllowDifferentMembersInActual` was removed for 1.9.20 for all targets.

mvicsokolova: Oh, ok

fzhinkin: Well, `AllowDifferent....` was actually removed from 1.9.20, but without the supression the warning is still here:
```
w: file:///Users/Filipp.Zhinkin/Development/kotlinx-atomicfu/atomicfu/src/jsAndWasmSharedMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt:5:25 Actual typealias 'SynchronizedObject': actual class and its non-final expect class must declare exactly the same non-private members. The following non-private members in actual class are mismatched:

property '_hashCode': the visibility of this member must be the same in the expect class and the actual class. This error happens because the expect class 'SynchronizedObject' is non-final

property 'typeInfo': the visibility of this member must be the same in the expect class and the actual class. This error happens because the expect class 'SynchronizedObject' is non-final
```

fzhinkin: * and the ID of the warning seems to have changed since pre-release 1.9.20 builds, that's why I proposed the fix above.

mvicsokolova: Ok, thank you! I just didn't see the warning is still there

PR #336, opened 2023-08-15
Update native atomics implementation

mvicsokolova:
This commit replaces deprecated native atomics from kotlin.native.concurrent package with new atomics from kotlin.concurrent, because deprecation levels of native atomics were updated for 1.9.20 (see KT-58123)

This commit was only present in the `kotln-community/dev` branch, now moving it to `develop`.


mvicsokolova:
It's better to merge these changes as 2 separate commits, so it would be easier to track these changes later.

woainikk (approved):

PR #344, opened 2023-09-07
Get rid of `previous-compilation-data.bin` file in META-INF

mvicsokolova:
`previous-compilation-data.bin` is Gradle's internal machinery file that was packed in the resulting artifact because we depended directly on the output of `compileJavaModuleInfo` task, depending on `compileJavaModuleInfo.destinationDirectory` solves the problem.

Similar problems:
https://github.com/Kotlin/kotlinx-datetime/issues/304
https://github.com/Kotlin/kotlinx.coroutines/issues/3668


integration-testing/build.gradle
```diff
@@ -0,0 +1,89 @@
+
+buildscript {
```
ilya-g, line 2: How do you intend to run this project? Will you set up another build on TC?

ilya-g (approved):

integration-testing/build.gradle
mvicsokolova: I'll add the second step in the build on TC that will depend on `publishToMavenLocal`

Nek-12:
When will a 0.23.0 be published? This is causing https://github.com/Kotlin/kotlinx.coroutines/issues/3668 to occur despite using 1.7.3 as a coroutines dependency

mvicsokolova:
Hi! I'm publishing 0.23.0 with this fix today

PR #345, opened 2023-09-11
Integration tests

mvicsokolova:
This PR introduces the framework for atomicfu-gradle-plugin integration tests.
To run functional tests:
```
cd integration-testing
./gradlew functionalTest
```

Current problems:

- [x] Do not copy versions in build scripts of sample projects #gradle
- [x] Add tests for the JS IR backend
- [x] `DependenciesChecker`: should also check apiElements and runtimeElements configurations
- [x] `DependenciesChecker`: it'd be nice to check .module or .pom files after publication of the project for "consumable" dependencies that will not be present in the `dependencies` output
- [x] `AtomicfuGradlePlugin`: configuration of dependencies for MPP project should be reviewed, currently it causes Mpp dependency test to fail

- [ ] klib dump-ir checks are not enabled for now because of this known problem [KT-61143](https://youtrack.jetbrains.com/issue/KT-61143), looking for WA


integration-testing/examples/jvm-sample/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+        mavenLocal()
+    }
+
+    dependencies {
+        classpath("org.jetbrains.kotlinx:atomicfu-gradle-plugin:$atomicfu_version")
+    }
+}
+
+plugins {
+    kotlin("jvm") version "1.9.10" // todo get kotlin_version from gradle.properties
```
mvicsokolova, line 14: What is a proper way to get `kotlin_version` from `gradle.properties` in build.gradle.kts files?

integration-testing/examples/jvm-sample/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
+buildscript {
+    val atomicfu_version = rootProject.properties["atomicfu_version"]
```
mvicsokolova, line 2: And what is the proper way to get atomicfu_version here?)

integration-testing/src/test/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Environment.kt
```diff
@@ -0,0 +1,20 @@
...
+package kotlinx.atomicfu.gradle.plugin.test.framework.runner
+
+import java.io.File
+
+internal const val ENABLE_JVM_IR_TRANSFORMATION = "kotlinx.atomicfu.enableJvmIrTransformation"
+internal const val ENABLE_JS_IR_TRANSFORMATION = "kotlinx.atomicfu.enableJsIrTransformation"
+internal const val ENABLE_NATIVE_IR_TRANSFORMATION = "kotlinx.atomicfu.enableNativeIrTransformation"
+
+// todo: get from System.properties
+internal val atomicfuVersion = "0.22.0-SNAPSHOT"
```
mvicsokolova, line 14: Can we somehow get `atomicfuVersion` from `integration-testing/gradle.properties`, maybe as a System property

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -85,23 +105,34 @@ private fun Project.configureDependencies() {
...
+        addJsCompilerPluginRuntimeDependency()
         configureMultiplatformPluginDependencies(version)
     }
 }
 
-private fun Project.configureTasks() {
-    val config = config
-    withPluginWhenEvaluated("kotlin") {
-        if (config.transformJvm) configureJvmTransformation()
+private fun Project.configureMultiplatformPluginDependencies(version: String) {
```
mvicsokolova, line 116: The only conceptual change in the `AtomicfuGradlePlugin` is in `configureMultiplatformPluginDependencies`, all other changes are just refactoring: changing the order of declarations.

mvicsokolova, line 116: The only conceptual change in the `AtomicfuGradlePlugin` is in `configureMultiplatformPluginDependencies`, all other changes are just refactoring: changing the order of declarations.

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
```diff
@@ -0,0 +1,83 @@
...
+    override fun clear() { targetDir.deleteRecursively() }
+}
+
+internal class BuildResult(exitCode: Int, private val logFile: File) {
+    val isSuccessful: Boolean = exitCode == 0
+
+    val output: String by lazy { logFile.readText() }
+
+    // Gets the list of dependencies for the given configuration
+    fun getDependenciesForConfig(configuration: String): List<String> {
```
fzhinkin, line 58: It deserves a unit test.

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
```diff
@@ -0,0 +1,83 @@
...
+    override val targetDir: File
+) : GradleBuild {
+
+    override var enableJvmIrTransformation = false
+    override var enableJsIrTransformation = false
+    override var enableNativeIrTransformation = false
+
+    private val properties
+        get() = buildList {
+            add("-P$ENABLE_JVM_IR_TRANSFORMATION=$enableJvmIrTransformation")
```
fzhinkin, line 37: Unless there is an intention to have particular `kotlin_version` and `atomicfu_version` values in examples' properties, it's better to set them here as well. Otherwise, we're doomed to update it with every atomicfu release or Kotlin version update.

integration-testing/build.gradle.kts
```diff
@@ -0,0 +1,113 @@
...
+    if (using_snapshot_version) {
+        repositories {
+            mavenLocal()
+            maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
+        }
+    }
+}
+
+plugins {
+    kotlin("jvm") version "${project.properties["kotlin_version"]}" // todo: is there a more consise way to get kotlin_version from gradle.properties?
```
fzhinkin, line 41: Maybe [versions catalog](https://developer.android.com/build/migrate-to-catalogs)?

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import kotlinx.atomicfu.gradle.plugin.test.framework.checker.*
+import kotlinx.atomicfu.gradle.plugin.test.framework.runner.*
+import kotlin.test.Test
+
+class MppProjectTest {
+    private val mppSample: GradleBuild = createGradleBuildFromSources("mpp-sample")
+
+    @Test
+    fun testMppJvm1() {
```
fzhinkin, line 15: Could you please use more verbose names (instead of `1` and `2`)?

integration-testing/examples/mpp-sample/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+kotlin {
+    jvm()
+    js()
+
+    val hostOs = System.getProperty("os.name")
+    val isArm64 = System.getProperty("os.arch") == "aarch64"
+    val isMingwX64 = hostOs.startsWith("Windows")
+    val nativeTarget = when {
```
fzhinkin, line 41: Does it make sense to enable all native targets (once it'll be possible to check them)?

integration-testing/build.gradle.kts
```diff
@@ -0,0 +1,113 @@
+import org.jetbrains.kotlin.gradle.utils.NativeCompilerDownloader
```
fzhinkin, line 1: Is it possible to make integration tests module a submodule of the main project? Currently, it's not that convenient to work with integration tests and it's easy to forget about running them locally as tests declared here will not be executed along with other project tests.

fzhinkin (changes requested):
Great job! Overall looks good, there's only a few minor issues to address.

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Utils.kt
```diff
@@ -0,0 +1,34 @@
...
+    val builder = ProcessBuilder(gradleCommands)
+    builder.directory(gradleWrapperDir)
+    builder.redirectErrorStream(true)
+    builder.redirectOutput(logFile)
+    val process = builder.start()
+    val exitCode = process.waitFor()
+    return BuildResult(exitCode, logFile)
+}
+
+private fun buildSystemCommand(projectDir: File, commands: List<String>, properties: List<String>): List<String> {
```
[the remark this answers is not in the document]
mvicsokolova: Probably, we can apply it later, it was postponed for now because of side effects of plugin application to sample projects.
@shanshin could you please elaborate on the details? 🙂

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
mvicsokolova: Right, yeah, we can set these versions manually here, or just set them in `integration-testing/gradle.properties` file to the same values as in the root gradle.properties

integration-testing/build.gradle.kts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+repositories {
+    mavenLocal()
+    maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
+    mavenCentral()
+}
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_11
```
shanshin, line 51: As far as I know, Gradle [recommends using toolchain block](https://docs.gradle.org/current/userguide/java_plugin.html#toolchain_and_compatibility)

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -85,23 +105,34 @@ private fun Project.configureDependencies() {
...
+    val multiplatformExtension = kotlinExtension as? KotlinMultiplatformExtension ?: error("Expected kotlin multiplatform extension")
+    val atomicfuDependency = "org.jetbrains.kotlinx:atomicfu:$version"
+    multiplatformExtension.sourceSets.getByName("commonMain").dependencies {
```
shanshin, line 119: Can there be a situation when there are no `commonMain` or `commonTest` source sets in the project?

integration-testing/examples/jvm-sample/gradle.properties
```diff
@@ -0,0 +1,2 @@
+kotlin_version=1.9.0
+atomicfu_version=0.22.0-SNAPSHOT
```
shanshin, line 2: Also, line wrapping here

integration-testing/examples/mpp-sample/build.gradle.kts
shanshin: @mvicsokolova, is this done so that the names of the native target are always the same?

integration-testing/build.gradle.kts
fzhinkin: `integration-testing` is not included into `atomicfu` Gradle project, one have to manually `integration-testing` and run tests from it.

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/checker/ArtifactChecker.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    val buildDir
+        get() = targetDir.resolve("build").also {
+            require(it.exists() && it.isDirectory) { "Could not find `build/` directory in the target directory of the project $projectName: ${targetDir.path}" }
+        }
+
+    abstract fun checkReferences()
+
+    protected fun ByteArray.findAtomicfuRef(): Boolean {
+        loop@for (i in 0 until this.size - ATOMIC_FU_REF.size) {
```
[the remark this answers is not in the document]
fzhinkin: It also wound be nice to have a unit test for it (as well as for https://github.com/Kotlin/kotlinx-atomicfu/pull/345#discussion_r1369766743)

integration-testing/build.gradle.kts
shanshin: I think in this case it is better to make a [composite build](https://docs.gradle.org/current/userguide/composite_builds.html) and run functional tests [when calling `check`](https://docs.gradle.org/current/userguide/composite_builds.html#included_build_task_dependencies) of the main project.

mvicsokolova: We need this separate module to be able to run sample project builds completely independently, like they are honest isolated user projects 👀
And also, in integration-testing I use locally published current atomicfu version.

fzhinkin: But the sample projects will be independent either way as for them we invoke Gradle manually and these runs (or builds, I'm not sure about correct wording) will not depend on surrounding project (after all, all sample project files are copied into temp directory, aren't they?).

integration-testing/examples/jvm-sample/build.gradle.kts
mvicsokolova: The error is: `The root project is not yet available for build.`
And I get the same error for version catalogs, so, I only managed to do it like this:

```
    versionCatalogs {
        create("libs") {
            version("atomicfu", "0.22.0-SNAPSHOT")
            version("kotlin", "1.9.20")
        }
    }
}
```

But it's still versions hardcoding 👀

mvicsokolova: I guess it's fine to leave like it is for now then

integration-testing/build.gradle.kts
mvicsokolova: When I configure version catalog like this, I get "The root project is not yet available for build."  👀

```
    versionCatalogs {
        create("libs") {
            version("atomicfu", gradle.rootProject.properties["atomicfu_version"])
        }
   }
}
```

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
mvicsokolova: Done

integration-testing/examples/mpp-sample/build.gradle.kts
mvicsokolova: That is just part of the default IDEA mpp project configuration. I guess, we need this to be able to call gradle build for this sample project on different targets on TC, right?

integration-testing/build.gradle.kts
mvicsokolova: Yes, you're right, project builds will be independent anyway. But I need to build them with a locally published atomicfu version, and I don't know how to make project tests depend on pubslishToMavenLocal task.

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Commands.kt
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2016-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.atomicfu.gradle.plugin.test.framework.runner
+
+internal fun GradleBuild.clean(): BuildResult = runGradle(listOf("clean"))
+
+internal fun GradleBuild.build(): BuildResult = runGradle(listOf("clean", "build"))
```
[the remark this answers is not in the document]
mvicsokolova: Fixed

integration-testing/examples/mpp-sample/build.gradle.kts
fzhinkin: All native targets could be enabled simultaneously and only those supported by the host compiler will be built.

integration-testing/build.gradle.kts
fzhinkin: Something like the following should work:
```
task checkAfterPublish(dependsOn: ':atomicfu-gradle-plugin:publishToMavenLocal')

tasks.named("check").configure {
    it.dependsOn(checkAfterPublish)
}
```

integration-testing/examples/mpp-sample/build.gradle.kts
mvicsokolova: Added all the native targets

mvicsokolova:
The current problem of the build failure: mavenTest depends on `atomicfu-gradle-plugin:publishToMavenLocal` now and the locally published  atomicfu-jvm:0.22.0-SNAPSHOT.jar should be added to it's classpath.

fzhinkin:
> The reason of the current build failure: mavenTest depends on `atomicfu-gradle-plugin:publishToMavenLocal` now and the locally published atomicfu-jvm:0.22.0-SNAPSHOT.jar should be added to it's classpath.

@mvicsokolova it seems like you also need to run `:atomicfu:publishToMavenLocal` before integration-testing's `compile*` tasks.

fzhinkin:
As an extra effort to keep things nice and clean, it might be worth adding a separate set of publishToLocal tasks that will write artifacts to a temporary repo in the build directory (like int tests do).
In that case, we can be sure that atomicfu-artifacts won't be published even to the local maven repo if something is broken.
Not sure how difficult it would be to add such "publishToBuildLocal" publication tasks though.

mvicsokolova:
I fixed `mavenTest` task by depending directly to the `:atomicfu:publishToMavenLocal` task, which is should be avoided, but I guess we can stick to this solution for now. I've created an issue to collect improvements for integration tests, including adding separate tasks to publish artifacts to the custom local repo. (#362)

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/checker/DependenciesChecker.kt
```diff
@@ -0,0 +1,72 @@
...
+    var index = 0
+    while (index < lines.size) {
+        val line = lines[index]
+        if (line.contains("\"group\": \"org.jetbrains.kotlinx\"") &&
+            lines[index + 1].contains("\"module\": \"atomicfu")) {
+            error("org.jetbrains.kotlinx.atomicfu dependency found in the .module file ${moduleFile.path}")
+        }
+        index++
+    }
+}
```
fzhinkin, line 72: newline, please

and in few other files where it's also missing

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
```diff
@@ -0,0 +1,63 @@
...
+        while (index < lines.size) {
+            val line = lines[index++]
+            if (line.takeWhile { it.isLetter() } == configuration) break
+        }
+        while(index < lines.size) {
+            val line = lines[index++]
+            if (line.isBlank() || line == "No dependencies") break
+            // trim leading indentations (\---) and symbols in the end (*):
+            // \--- org.jetbrains.kotlinx:atomicfu:0.22.0-SNAPSHOT (n)
+            result.add(line.dropWhile { !it.isLetter() }.substringBefore(" "))
```
fzhinkin, line 51: A group name may start with a digit (although it's against the Maven Central naming guideline).

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
```diff
@@ -0,0 +1,63 @@
...
+    val output: String by lazy { logFile.readText() }
+
+    // Gets the list of dependencies for the given configuration
+    fun getDependenciesForConfig(configuration: String): List<String> {
+        val lines = output.lines()
+        val result = mutableListOf<String>()
+        var index = 0
+        while (index < lines.size) {
+            val line = lines[index++]
+            if (line.takeWhile { it.isLetter() } == configuration) break
```
fzhinkin, line 44: That will work incorrectly if someone is looking for a configuration named "blablabla", but there's a configuration named "blablabla42" (yes, I understand that it's not a problem right now:).

Does it make sense to check that a line starts with `"$configuration - "` substring instead?

fzhinkin (changes requested):

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
mvicsokolova: Fixed this)

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
mvicsokolova: Ya, you're right, anything can happen to configuration names) Fixed.

mvicsokolova: Though, it's not always proceeded with `"configuration - "`
So, replaced with:
`line.substringBefore(" ") == configuration`

integration-testing/examples/mpp-sample/src/commonTest/kotlin/IntArithmeticTest.kt
```diff
@@ -0,0 +1,15 @@
...
+
+class IntArithmeticTest {
+
+    @Test
+    fun testInt() {
+        val a = IntArithmetic()
+        a.doWork(1234)
+        assertEquals(1234, a.x)
+    }
+}
```
fzhinkin, line 15:
```suggestion
}

```

integration-testing/build.gradle.kts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+val mavenTest by tasks.registering(Test::class) {
+    testClassesDirs = sourceSets["mavenTest"].output.classesDirs
+    classpath = sourceSets["mavenTest"].runtimeClasspath
+    
+    dependsOn(":atomicfu:publishToMavenLocal")
+}
+
+val functionalTest by tasks.registering(Test::class) {
```
fzhinkin, line 61: It seems like there's a problem with the task result's caching.

What I did locally:
- checked out this PR;
- executed `:integration-testing:check`, all tests passed
- broke `AtomicFUGradlePlugin` by commenting out dependencies inclusion;
- executed `:integration-testing:check` once again and this time, there were no failures too.
Only by performing `clean` and then running the `check` again I was able to get test failures.

I think we can plug `outputs.upToDateWhen { false }` for now and address the issue later in https://github.com/Kotlin/kotlinx-atomicfu/issues/362

mvicsokolova: Thank you, I didn't think of that
Fixed for both mavenTest and functionalTest, and mentioned the problem in #362

mvicsokolova:
Rebased on the latest develop with Kotlin updated to 1.9.20

fzhinkin (approved):

mvicsokolova:
Well, something happened and mpp-sample build is failing with this error again 🪦
```
> Task :compileKotlinJs FAILED
e: java.lang.AssertionError: Built-in class kotlin.Any is not found
  at org.jetbrains.kotlin.builtins.KotlinBuiltIns$3.invoke(KotlinBuiltIns.java:93)
  at org.jetbrains.kotlin.builtins.KotlinBuiltIns$3.invoke(KotlinBuiltIns.java:88)
```

integration-testing/examples/mpp-sample/build.gradle.kts
```diff
@@ -25,6 +25,7 @@ apply(plugin = "kotlinx-atomicfu")
 
 repositories {
     mavenCentral()
+    maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
```
fzhinkin, line 28: Is there a reason to have these repos ordered differently in jvm- and mpp-sample projects?

mvicsokolova: The problem with unresolved classes from `kotlin-stdlib` may've been caused by the maven local repository being corrupted by some other process, and it's priority caused gradle to search for stdlib classes there first. As gradle can not work properly with `mavenLocal`, it's better not to set it first in the repo list .

PR #347, opened 2023-09-13
Sign publication failure again

mvicsokolova:
This issue was revealed after enabling integration testing of atomicfu. Now the first TC build step executes `build` and `publishToMavenLocal` tasks. `publishToMavenLocal` task fails with an error: 
```
Reason: Task ':atomicfu:compileTestKotlinLinuxX64' uses this output of task ':atomicfu:signLinuxX64Publication' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

This is an already known problem [KT-61313](https://youtrack.jetbrains.com/issue/KT-61313) caused by the task configuration performed by KGP.

This fix is necessary to be able to launch integration tests added in this PR: #344

### DIFF
--- a/integration-testing/build.gradle
+++ b/integration-testing/build.gradle
@@ -50,12 +50,8 @@ repositories {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_11
-    targetCompatibility = JavaVersion.VERSION_11
-}
-
-kotlin {
-    jvmToolchain(11)
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
 }
 
 dependencies {
@@ -80,4 +76,14 @@ task mavenTest(type: Test) {
     def sourceSet = sourceSets.mavenTest
     testClassesDirs = sourceSet.output.classesDirs
     classpath = sourceSet.runtimeClasspath
+}
+
+compileTestKotlin {
+    kotlinOptions.jvmTarget = "1.8"
+}
+
+compileKotlin {
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
 }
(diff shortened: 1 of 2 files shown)


integration-testing/build.gradle
```diff
@@ -50,8 +50,12 @@ repositories {
...
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility = JavaVersion.VERSION_11
+    targetCompatibility = JavaVersion.VERSION_11
+}
+
+kotlin {
+    jvmToolchain(11)
```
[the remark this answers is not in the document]
mvicsokolova: Fixed this, I'll commit that change separately

integration-testing/build.gradle
```diff
@@ -50,12 +50,8 @@ repositories {
...
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_11
-    targetCompatibility = JavaVersion.VERSION_11
-}
-
-kotlin {
-    jvmToolchain(11)
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
```
ilya-g, line 54: Hmm, still don't quite get why there's a change here. Is something changed in the base branch?

ilya-g: Also the build is failing likely because of this change

mvicsokolova: My bad :(( Just reverted the commit and didn't notice, that old jdk 8 targets are still there. Fixed it now.

ilya-g (approved):

ilya-g:
Btw, regarding 'clean' task failure on Windows, you can take a look how it is worked around in kotlinx-datetime.

Issue #370, opened 2023-11-23
The problem of 2 atomicfu gradle plugins


mvicsokolova:
There are 2 atomicfu gradle plugins: `kotlinx.atomicfu:atomicfu-gradle-plugin` provided by the library and `kotlin:atomicfu` compiler plugin, the latter is not supposed to be applied directly and is applied by the `atomicfu-gradle-plugin`.

They not only create confusion among users but also serve as the cause of an important issue: `atomicfu-gradle-plugin`, when applied to the user project, should apply compiler plugin of the kotlin version used in the user project, not the kotlin version that was used to build `atomicfu-gradle-plugin` itself. That is not the case currently.

**The reason:** compiler plugin is found in the classpath of the user project, because it is stated as the [implementation](https://github.com/Kotlin/kotlinx-atomicfu/blob/f83c00d1492c83cfd90bbb3c1607412f92244078/atomicfu-gradle-plugin/build.gradle#L26) dependency of the `atomicfu-gradle-plugin` and hence leaks into the user project transitively, though with the kotlin version used in kotlinx.atomicfu.

**The fast solution:** for now the behaviour is correct when kotlin versions used in a user project and kotlinx.atomicfu are equal. Otherwise, `atomicfu-gradle-plugin` should ask a user to add the compiler plugin with the kotlin version used in the project directly:

```
buildscript {
    repositories {
        mavenCentral()
    }

    dependencies {
        classpath("org.jetbrains.kotlinx:atomicfu-gradle-plugin:$atomicfu_version}")
        classpath("org.jetbrains.kotlin:atomicfu:$my_kotlin_version")
    }
}
```

**The right solution**: getting rid of the gradle plugin on the side of the compiler and leaving only one gradle plugin in the library may help. To be further investigated.

mvicsokolova:
I see 2 possible options to address this problem: 

> I'll further refer to the plugins as the **_compiler_** plugin (AtomicfuGradleSubplugin in kotlin repo) and the **_gradle_** plugin (AtomicfuGradlePlugin in kotlinx-atomicfu repo). 

**1 Solution**. Remove the **_compiler_** plugin from kotlin repo and stick to the **_gradle_** plugin:

**Pros:**
- The release cycle will not be tied to Kotlin releases, hot-fixes may be published quickly if necessary. 
[...]

Issue #362, opened 2023-11-08
Enhancements for integration testing


mvicsokolova:
- [ ] check Native and JS/IR klibs for atomicfu references (see [KT-61143](https://youtrack.jetbrains.com/issue/KT-61143/klib-tool-Dump-IR-even-if-there-are-interop-dependencies)).

- [ ] `mavenTest` task directly depends on `:atomicfu` local publication task, which should be avoided. Create separate tasks for publication of `atomicfu` and `atomicfu-gradle-plugin` artifacts to the custom local repo in the build directory.

PR #377, opened 2023-12-06
Apply atomicfu compiler plugin directly in buildscript classpath.

mvicsokolova:
* Atomicfu compiler plugin should be manually added to the classpath by the user, otherwise the error is thrown by the plugin
* All kotlin-stdlib transitive dependencies that leaked into user projects are removed

This commit is part of the solution to the problem described in #370


mvicsokolova:
I've added a separate commit with removal of old atomicfu-gradle-plugin tests to these changes.

- Those tests did not honestly publish the current versin of atomicfu-gradle-plugin to the local repo
- The current integration tests cover their functionality

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -45,12 +45,35 @@ open class AtomicFUGradlePlugin : Plugin<Project> {
...
         configureDependencies()
         configureTasks()
     }
 }
 
+private fun Project.checkClasspathForAtomicfuCompilerPlugin(pluginVersion: String?) {
+    val kotlinVersion = getKotlinPluginVersion()
+    rootProject.buildscript.configurations.findByName("classpath")
+        ?.allDependencies?.find { it.group == "org.jetbrains.kotlin" && it.name == "atomicfu" }
+        ?: error("Please add a dependency to the atomicfu compiler plugin in the buildscript classpath configuration, " +
```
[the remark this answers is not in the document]
mvicsokolova: Updated the README

integration-testing/examples/jvm-sample/build.gradle.kts
```diff
@@ -1,10 +1,14 @@
 buildscript {
     repositories {
         mavenLocal()
```
shanshin, line 3: It seems to me that in the future it is better to use custom repositories for functional tests, rather than `mavenLocal()`

shanshin (approved):

integration-testing/examples/jvm-sample/build.gradle.kts
mvicsokolova: + I've added that to the ticket: https://github.com/Kotlin/kotlinx-atomicfu/issues/362

fzhinkin (approved):

PR #380, opened 2023-12-07
Always add a transitive atomicfu dependency for native targets.

### LINKED ISSUE #379: Could not find "org.jetbrains.kotlinx:atomicfu-cinterop-interop" when using coroutines 1.8.0-RC
The following error was reproduced in projects, that depended solely on kotlinx.coroutines 1.8.0-RC and did not have a direct dependency to atomicfu: 

```
> Task :compileKotlinIosArm64 FAILED
e: Could not find "org.jetbrains.kotlinx:atomicfu-cinterop-interop" in [/.konan/klib,/.konan/kotlin-native-prebuilt-macos-aarch64-1.9.21/klib/common, /.konan/kotlin-native-prebuilt-macos-aarch64-1.9.21/klib/platform/ios_arm64]
```

**The reason:** `kotlinx.coroutines` were built with the latest version of `atomicfu-gradle-plugin` that does not provide a transitive dependency to atomicfu in case of enabled Native IR transformation. Though it is still required, because coroutines use declarations from `kotlinx.atomicfu.locks` module. 

That is a problem on kotlinx.atomicfu side, it should always add the implementation atomicfu dependency for native targets.

The WA is adding atomicfu dependency manually: 

```
val nativeMain by creating {
  dependencies {
    implementation("org.jetbrains.kotlinx:atomicfu:0.23.1")
  }
}
```

Also, see: https://github.com/Kotlin/kotlinx.coroutines/issues/3968

mvicsokolova:
Fixes #379

Can be tested by building integration tests in kotlinx.coroutines (https://github.com/Kotlin/kotlinx.coroutines/pull/3974).

### DIFF
--- a/integration-testing/examples/user-project/build.gradle.kts
+++ b/integration-testing/examples/user-project/build.gradle.kts
@@ -40,3 +40,9 @@ kotlin {
         commonTest {}
     }
 }
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+    kotlinOptions {
+        freeCompilerArgs += listOf("-Xskip-prerelease-check")
+    }
+}
--- a/integration-testing/examples/user-project/build.gradle.kts
+++ b/integration-testing/examples/user-project/build.gradle.kts
@@ -8,7 +8,10 @@ plugins {
 repositories {
     mavenCentral()
     maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
-    maven("../mpp-sample/build/.m2/")
+    val localRepositoryUrl = findProperty("localRepositoryUrl")
+    if (localRepositoryUrl != null) {
+        maven(localRepositoryUrl)   
+    }
     mavenLocal()
 }
 
--- a/integration-testing/examples/user-project/build.gradle.kts
+++ b/integration-testing/examples/user-project/build.gradle.kts
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2016-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+plugins {
+    kotlin("multiplatform") version libs.versions.kotlinVersion.get()
+}
+
+repositories {
+    mavenCentral()
+    maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
+    maven("../mpp-sample/build/.m2/")
+    mavenLocal()
+}
+
+kotlin {
+    jvm()
+
+    js()
+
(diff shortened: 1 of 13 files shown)


fzhinkin:
That's nice (although not sure for whom) to have a test in a separate project that could catch issues in atomicfu, but our own integration tests should be capable of catching the same issue. Please add the missing test cases.

mvicsokolova:
In the [module](https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core/1.8.0-RC/kotlinx-coroutines-core-1.8.0-RC.module) file of kotlinx.coroutines 1.8.0-RC atomicfu dependency was not declared: 
```
"dependencies": [
        {
          "group": "org.jetbrains.kotlin",
          "module": "kotlin-stdlib",
          "version": {
            "requires": "1.9.21"
          }
        }
      ],
```

I locally published atomicfu with this fix -> built and published coroutines. The coroutines klib contains the atomicfu dependency now:
```
"dependencies": [
        {
          "group": "org.jetbrains.kotlin",
          "module": "kotlin-stdlib",
          "version": {
            "requires": "1.9.21"
          }
        },
        {
          "group": "org.jetbrains.kotlinx",
          "module": "atomicfu",
          "version": {
            "requires": "0.23.1-SNAPSHOT-transitive-atomicfu"
          }
        }
      ],
```

integration-testing/examples/user-project/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright 2016-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+plugins {
+    kotlin("multiplatform") version libs.versions.kotlinVersion.get()
+}
+
+repositories {
+    mavenCentral()
+    maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
+    maven("../mpp-sample/build/.m2/")
```
mvicsokolova, line 11: So, for now this solution works only for manual publishing of mpp-sample.

1. publishToMavenLocal atomicfu 0.23.1-SNAPSHOT

publish mpp-sample library to `mpp-sample/build/.m2/...`
2. `cd integration-testing/examples/mpp-sample/  gradle clean publish `
build user-project
3. `cd ../user-project/ gradle clean build `

And to write a test, I need to publish mpp-sample library to some known directory, so that I can provide a path to the artefact in user-project. 
What would be the best way to do that? 🤔

mvicsokolova:
Rebased on develop and added `-Xskip-prelease-check` flag to the user-project.

PR #386, opened 2023-12-20
Alternative solution how to add compiler plugin dependency to the project

mvicsokolova:
Add a transitive dependency to the compiler plugin in atomicfu-gradle-plugin.

Do not require to add it manually.

Problem: not sure that compiler plugin dependency will be resolved correctly on the user side.


atomicfu-gradle-plugin/build.gradle
```diff
@@ -24,76 +24,12 @@ dependencies {
     // Atomicfu compiler plugin dependency will be loaded to kotlinCompilerPluginClasspath
     // Atomicfu plugin will only be applied if the flag is set kotlinx.atomicfu.enableJsIrTransformation=true
     compileOnly "org.jetbrains.kotlin:atomicfu:$kotlin_version"
+    runtimeOnly "org.jetbrains.kotlin:atomicfu:please-add-dependency-in-your-project"
```
fzhinkin, line 27: Maybe it's worth specifying some min viable, but existing version? I guess, more up-to-date version, if presented, should be preferred to it during dependencies resolution, but it also should help with all those failure in CI.

mvicsokolova: Right, this seems like the only solution. Otherwise I can not avoid `atomicfu-gradle-plugin:runtimeClasspath` resolution.

mvicsokolova:
- I've added runtimeOnly `"org.jetbrains.kotlin:atomicfu:1.7.0"` dependency to :atomicfu-gradle-plugin
- Removed manual addition of the compiler plugin to the classpath of user projects (in jvm-sample, mpp-sample)
- `plugin-order-bug` test now builds successfuly

atomicfu-gradle-plugin/build.gradle
```diff
@@ -24,76 +24,12 @@ dependencies {
     // Atomicfu compiler plugin dependency will be loaded to kotlinCompilerPluginClasspath
     // Atomicfu plugin will only be applied if the flag is set kotlinx.atomicfu.enableJsIrTransformation=true
     compileOnly "org.jetbrains.kotlin:atomicfu:$kotlin_version"
+    runtimeOnly "org.jetbrains.kotlin:atomicfu:1.7.0"
```
fzhinkin, line 27: Please add the comment describing why we have two different dependencies on two different version of the same artifact.
It would be also nice to describe why you chose 1.7.0 instead of, lets say, 1.6.20.

fzhinkin (changes requested):

atomicfu-gradle-plugin/build.gradle
mvicsokolova: Done)

fzhinkin: Thanks!

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/PluginOrderBugTest.kt
```diff
@@ -17,4 +18,14 @@ class PluginOrderBugTest {
         val buildResult = pluginOrderBugProject.cleanAndBuild()
         assertTrue(buildResult.isSuccessful, buildResult.output)
     }
+    
+    @Test
+    fun testResolvedCompilerPluginDependency() {
```
mvicsokolova, line 23: I've added some check that there is `org.jetbrains.kotlin:atomicfu:1.9.20` dependency in the project classpath.

In the buildEnvironment we can also find the resolution string: 
`org.jetbrains.kotlin:atomicfu:1.6.20 -> 1.9.20`

but probably just checking for the kotlin version is enough

fzhinkin: Thanks to covering such a scenario!

mvicsokolova:
Rebased on the latest develop with increased Metaspace

mvicsokolova:
Rebased on the latest develop with increased Metaspace

mvicsokolova:
Maybe we can postpone solution of the problem with the compileOnly atomicfu dependency in Native targets for now?
Seems, that it requires more discussion and there is no quick solution. WDYT?

Though the resolution of this warning for kotlin-stdlib dependency seems ok for now.

fzhinkin (approved):

PR #391, opened 2024-01-03
Resolve build warnings

mvicsokolova:
In this PR I've resolved some build warnings.

### DIFF
--- a/atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
+++ b/atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/AtomicFU.kt
@@ -77,8 +77,10 @@ public actual class AtomicBoolean internal constructor(@PublishedApi internal va
         get() = a.value != 0
         set(value) { a.value = if (value) 1 else 0 }
 
+    @InlineOnly
     public actual inline operator fun getValue(thisRef: Any?, property: KProperty<*>): Boolean = value
 
+    @InlineOnly
     public actual inline operator fun setValue(thisRef: Any?, property: KProperty<*>, value: Boolean) { this.value = value }
 
     public actual inline fun lazySet(value: Boolean) { this.value = value }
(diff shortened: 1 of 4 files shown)


mvicsokolova:
Currently, I'm trying to figure out what is the right way to add kotlin-stdlib dependency to atomicfu native targets.

There is this warning currently: 
```
A compileOnly dependency is used in the Kotlin/Native target 'linuxX64':
Compilation: main

Dependencies:
org.jetbrains.kotlin:kotlin-stdlib

Such dependencies are not applicable for Kotlin/Native, consider changing the dependency type to 'implementation' or 'api'.
To disable this warning, set the kotlin.native.ignoreIncorrectDependencies=true project property
```

Compile only dependencies are not applicable because dependencies are statically linked. There is no option to change a dependency at runtime.

On the other hand, AFAIK `kotlin-stdlib` dependency can not transitively leak into the user project, the user project should use `kotlin-stdlib` of the Kotlin version bundled in Gradle. That's why all `kotlin-stdlib` dependencies in `:atomicfu` were added as compileOnly.

gradle.properties
```diff
@@ -29,3 +29,5 @@ kotlin.mpp.enableCInteropCommonization=true
 # Workaround for Bintray treating .sha512 files as artifacts
 # https://github.com/gradle/gradle/issues/11412
 systemProp.org.gradle.internal.publish.checksums.insecure=true
+
+org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=2048m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
fzhinkin, line 33:
```suggestion
org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=2g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
Just for consistency.

Also, why do we need to specify file encoding explicitly? And why did everything work somehow without it before?

mvicsokolova: We don't need to set the encoding, I just read gradle docs and left the config from the example as it was. We don't need a heap dump either.

mvicsokolova:
Yes, fixed gradle.properties in all the integration tests as well

PR #392, opened 2024-01-09
Bobova/train fix

woainikk:


mvicsokolova:
Hi! Do we actually need mavenLocal()? Is custom Kotlin repository not enough?  👀

mvicsokolova (approved):

PR #394, opened 2024-01-18
Add all sources to the transformed directory

### LINKED ISSUE #388: After applying atomicfu plugin to a project with Java sources corresponding classes are missing
Based on https://github.com/Kotlin/kotlinx-atomicfu/issues/384#issuecomment-1869007976

If a project contains both Kotlin and Java sources, application of the `atomicfu` plugin leads to Java classes being lost.

Here's a reproducer: https://github.com/fzhinkin/atomicfu-issues-with-java-sources

Everything works fine with atomicfu version below `0.21.0`.

mvicsokolova:
In #303 we introduced new directories: `atomicfu-orig` for original sources and `atomicfu` for transformed sources. Though only transformed Kotlin sources were placed into the `atomicfu` directory, that caused a runtime error for projects with Java sources.

In this PR I copy both Kotlin and Java sources to the transformed directory.

In the previous [commit](https://github.com/Kotlin/kotlinx-atomicfu/commit/6c7ca4b03d465f9f2ac399e9f11ef4b6a312f1d0), that was supposed to fix this error, I only added Java sources to the transformed directory.

Fixes #388

### DIFF
--- a/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
+++ b/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
@@ -337,8 +337,9 @@ private fun Project.configureTransformationForTarget(target: KotlinTarget) {
                 it.destinationDirectory.value(originalDestinationDirectory)
             }
         }
-        val originalClassesDirs: FileCollection =
-            project.files(classesDirs.from.toTypedArray()).filter { it.exists() }
+        val originalClassesDirs: FileCollection = 
+            project.objects.fileCollection().from(compilationTask.flatMap { it.destinationDirectory }) + 
+                project.files(classesDirs.from.toTypedArray()).filter { it.exists() }
         originalDirsByCompilation[compilation] = originalClassesDirs
         val transformedClassesDir = project.layout.buildDirectory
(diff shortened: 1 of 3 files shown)


atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -337,8 +337,9 @@ private fun Project.configureTransformationForTarget(target: KotlinTarget) {
                 it.destinationDirectory.value(originalDestinationDirectory)
             }
         }
-        val originalClassesDirs: FileCollection =
-            project.files(classesDirs.from.toTypedArray()).filter { it.exists() }
+        val originalClassesDirs: FileCollection = 
+            project.objects.fileCollection().from(compilationTask.flatMap { it.destinationDirectory }) + 
+                project.files(classesDirs.from.toTypedArray()).filter { it.exists() }
```
Tapchicoma, line 342: Nit: You could call multiple times `.from` to add more things into `ConfigurableFileCollection`:

```suggestion
        val originalClassesDirs: FileCollection = project.objects.fileCollection()
            .from(compilationTask.flatMap { it.destinationDirectory })
            .from({ project.files(classesDirs.from).filter { it.exists() } })
```

Tapchicoma (approved):

Issue #376, opened 2023-12-05
Warning: compileOnly dependencies are not applicable for Kotlin/Native


mvicsokolova:
During the atomicfu build, different warnings are thrown. For example: 

```
A compileOnly dependency is used in the Kotlin/Native target 'linuxX64':
                        Compilation: main

                        Dependencies:
                        org.jetbrains.kotlin:kotlin-stdlib
org.jetbrains.kotlin:kotlin-stdlib-common

Such dependencies are not applicable for Kotlin/Native, consider changing the dependency type to 'implementation' or 'api'.
To disable this warning, set the kotlin.native.ignoreIncorrectDependencies=true project property
```

It'd be great to preliminary resolve them to prevent errors in the future.

lppedd:
Is this going to become a blocking issue when updating to Kotlin 2.0?

m-sasha:
Could you clarify whether this is fixed (by https://github.com/Kotlin/kotlinx-atomicfu/pull/395) or does it still require Kotlin 2.0?

PR #395, opened 2024-01-19
Add atomicfu to the apiElements config for native targets

mvicsokolova:
The warning about `compileOnly` dependencies declared for native sourceSets will be specified in 2.0 and will only be triggered when dependency is not present in the `apiElements` config of the native sourceSet. See [KT-64109](https://youtrack.jetbrains.com/issue/KT-64109) for more details.

The corresponding issue: #376

### DIFF
--- a/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
+++ b/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
@@ -128,9 +128,10 @@ private fun Project.configureMultiplatformPluginDependencies(version: String) {
         implementation(atomicfuDependency)
     }
     // Include atomicfu as a dependency for publication when transformation for the target is disabled
-    multiplatformExtension.targets.all { target ->
-        // Add an implementation dependency for native/wasm targets or if transformation is disabled
-        if (isTransitiveAtomicfuDependencyRequired(target)) {
+    multiplatformExtension.targets
+        .matching { target -> isTransitiveAtomicfuDependencyRequired(target) }
+        .all { target ->
+            // Add an implementation dependency for native/wasm targets or if transformation is disable
             target.compilations.all { compilation ->
                 compilation
                     .defaultSourceSet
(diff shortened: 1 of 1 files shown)


atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -129,6 +129,7 @@ private fun Project.configureMultiplatformPluginDependencies(version: String) {
     }
     // Include atomicfu as a dependency for publication when transformation for the target is disabled
     multiplatformExtension.targets.all { target ->
```
adam-enko, line 131: A related nitpick: typically `all {}` should be avoided, or where possible, limited in its impact by using `matching {}`. 

`configureEach {}` is preferred, because it doesn't trigger the realisation of all the contents of a Gradle collection. In contrast `all {}` will eagerly configure all elements. Sometimes `all {}` is required - and since Atomicfu is adding a dependency I think this is one such case. Without it, the dependency might not be added in time. (I don't have a hard or fast rule for this - Gradle is mysterious sometimes!)

Because the `all {}` block contains some conditional logic, it might well eagerly configure a target unnecessarily. The impact can be avoided by moving the `if` logic into a `matching {}` filter.

For this function because there are two different if statements, that means two different `all {}` blocks. Something like this:

```kotlin
// Include atomicfu as a dependency for publication when transformation for the target is disabled
multiplatformExtension.targets
  .matching { target -> isTransitiveAtomicfuDependencyRequired(target) }
  .all { target ->
    // Add an implementation dependency for native/wasm targets or if transformation is disable
        target.compilations.all { compilation ->
            compilation
                .defaultSourceSet
                .dependencies {
                    implementation(atomicfuDependency)
                }
        }
    }
  }

// atomicfu should also appear in apiElements config for native targets, 
// otherwise the warning is triggered, see: KT-64109
multiplatformExtension.targets
  .matching { target -> target.platformType == KotlinPlatformType.native }
  .all { target ->
    target.compilations.all { compilation ->
        compilation
            .defaultSourceSet
            .dependencies {
                api(atomicfuDependency)
            }
    }
}
```


However, this will be a behaviour change that you might not want to implement in this PR. I'll leave it up to you!

adam-enko (approved):
LGTM! 

I've got one suggestion for better Gradle compatibility that you might want to pick up in this commit, or consider for later.

One other question: do you have a test that shows this helps and avoids the "compileOnly not compatible with Kotlin/Native" warning message?

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
mvicsokolova: Thank you! That was a very useful comment. I think, I'll better schedule this change to the separate PR just for the clear history.

mvicsokolova:
Regarding the test to check whether we got rid of the warning: I still get it when I build the mpp project with a native target, but now because the compileOnly dependency is still included in the common sourceSet. And this will only be fixed in 2.0.

mvicsokolova:
Rebased on the latest develop.

Issue #210, opened 2021-11-16
document non-root project behavior for `org.jetbrains.kotlinx:atomicfu-gradle-plugin`


anatawa12:
According to the code below, if the `org.jetbrains.kotlinx:atomicfu-gradle-plugin` dependency is declared on a non-root project,  'kotlinx-atomicfu' Gradle Plugin can't detect the version of `atomicfu-gradle-plugin` and it doesn't add dependency automatically.

https://github.com/Kotlin/kotlinx.atomicfu/blob/de4e55674c51f58cbc1f16b097d68c4ba42a5e7b/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt#L30

Can you document this behavior?

dimsuz:
Oh, so that's why I'm getting an "unresolved reference androidfu", after applying it to a non-root project. Not obvious at all, should be documented indeed. But it really shouldn't depend on being added to a root project, this breaks modularization.

tadfisher:
A workaround is to add the plugin in `settings.gradle{.kts}` (in the `pluginManagement` block, for Gradle 7.+, with `apply false`) or in the root project's `plugins` block with `apply false`. Then it is added to the buildscript classpath without being applied to the root project.

Lukellmann:
Another option would be including a file with the current version in the plugin jar, see https://discuss.gradle.org/t/how-can-a-custom-gradle-plugin-determine-its-own-version/36761

mvicsokolova:
This issue will be fixed in the next release with these changes: #405

Now the version of the library is stored in the specified file, packed in the `atomicfu-gradle-plugin.jar`, and `AtomicfuGradlePlugin` gets the version of the applied plugin from this file, not from the root buildscript.

Issue #137, opened 2020-05-09
dependenciesVersion is null by default in AtomicFUPluginExtension


whyoleg:
When using gradle 6+ and adding `atomicfu-gradle-plugin` to classpath through `settings.gradle[.kts]` results in `pluginVersion=null` and as a result runtime dependency isn't added at all.
The problem is here:
https://github.com/Kotlin/kotlinx.atomicfu/blob/cffe6eeef32eec8a7dca0f7d414b7ae1b1306bd2/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt#L30
There are at least 2 workarounds:
- set version explicitly
- add `atomicfu-gradle-plugin` to classpath in root project
But it will be good if it will be fixed, so all classpathes for plugins could be configured inside `settings.gradle[.kts]`

fzhinkin:
Seems like it was resolved by https://github.com/Kotlin/kotlinx-atomicfu/pull/405

PR #405, opened 2024-02-27
Look for atomicfu-gradle-plugin version in the buildscript of the current project as well.

### LINKED ISSUE #403: atomicfu-gradle-plugin does not configure dependencies, when it applied in a non-root build sript.
There is a bug in resolution of `atomicfu-gradle-plugin` version, when atomicfu plugin is applied in the build script of the module, and not in the rootProject. In this case atomicfu does not configure necessary dependencies:  `compileOnly` library dependency for commonMain and `implementation` dependency for commonTest sourceSet, thus causing compilation errors.

mvicsokolova:
We only looked for `atomicfu-gradle-plugin` version in the buildscript of the rootProject, though atomicfu may be applied for one module only, in this case we failed to resolve it's version.

An example of the project: `integration-testing/examples/multi-module-test`.

Fixes #403


atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,15 +42,38 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
...
+                "```\n" +
+                "For further information, refer to the project's README: https://github.com/Kotlin/kotlinx-atomicfu/blob/master/README.md#apply-plugin"}
         extensions.add(EXTENSION_NAME, AtomicFUPluginExtension(pluginVersion))
         applyAtomicfuCompilerPlugin()
         configureDependencies()
         configureTasks()
     }
 }
 
+private fun Project.getAGPVersion(): String? =
```
fzhinkin, line 73: Sorry to say that, but the AGP abbreviation was already taken by the Android Gradle Plugin, so it worth using some other name here. 

It took me a while to understand that the problem you fixed has nothing to do with android. :)

mvicsokolova: Sorry, fixed that 🥲

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,15 +42,38 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
 open class AtomicFUGradlePlugin : Plugin<Project> {
     override fun apply(project: Project) = project.run {
         checkCompatibility()
-        val pluginVersion = rootProject.buildscript.configurations.findByName("classpath")
-            ?.allDependencies?.find { it.name == "atomicfu-gradle-plugin" }?.version
+        // Get the version of atomicfu-gradle-plugin applied to the project:
+        // first try to find it in the root buildscript, then in the buildscript of the current project.
+        // The error is thrown in case atomicfu-gradle-plugin is not applied.
+        val pluginVersion = rootProject.getAGPVersion() ?: this.getAGPVersion()
+        requireNotNull(pluginVersion) { "We were unable to find the version of `kotlinx-atomicfu` plugin applied to the project.\n" +
```
fzhinkin, line 49: Nice message!

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,15 +42,38 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
 open class AtomicFUGradlePlugin : Plugin<Project> {
     override fun apply(project: Project) = project.run {
         checkCompatibility()
-        val pluginVersion = rootProject.buildscript.configurations.findByName("classpath")
-            ?.allDependencies?.find { it.name == "atomicfu-gradle-plugin" }?.version
+        // Get the version of atomicfu-gradle-plugin applied to the project:
+        // first try to find it in the root buildscript, then in the buildscript of the current project.
+        // The error is thrown in case atomicfu-gradle-plugin is not applied.
+        val pluginVersion = rootProject.getAGPVersion() ?: this.getAGPVersion()
```
fzhinkin: Does it work for projects having deeper subprojects layout?
```
<root project>
   <subproject A>
     <subproject A.1>
     <subproject A.2>
   <subproject B>
```

mvicsokolova: Mm, probably, it will break if the current project is A.1 and `kotlinx-atomicfu` is applied in the buildscript of A.
Actually, we should walk through the whole project tree, I'm not sure, if there is an easier way to do that. 

I'll have a look 👀

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,15 +42,38 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
...
+                "Please ensure that `kotlinx-atomicfu` plugin is correctly applied.\n" +
+                "Below is an example of Kotlin DSL code that applies `kotlinx-atomicfu` plugin:\n" +
+                "```\n" +
+                "buildscript {\n" +
+                "    repositories {\n" +
+                "        mavenCentral()\n" +
+                "    }\n" +
+                "\n" +
+                "    dependencies {\n" +
+                "      classpath(\"org.jetbrains.kotlinx:atomicfu-gradle-plugin:0.23.2\")\n" +
```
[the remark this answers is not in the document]
whyoleg: > Should we take a version from project settings when building the plugin or something like that?

If you are gonna to decide to do this, then all this hacks to fetch an atomicfu version from a build script will be not needed at all, as you will already know a version of the plugin :)

And IMO, this could be a better approach, that introducing this good error message, because in that case, there will be no need for it as there will be no need to add `atomicfu` to build classpath, and so just applying plugin will be enough :)
It will also resolve issues like #210, #137 or any other issues when users **just** apply plugin in some way (f.e via version catalogs) and not via `buildscript.classpath` (which is, AFAIK, is Gradle legacy and is not suggested approach)

There are multiple ways to do it, from [manual ones](https://discuss.gradle.org/t/how-can-a-custom-gradle-plugin-determine-its-own-version/36761/2) to some existing Gradle plugins like [this](https://github.com/gmazzo/gradle-buildconfig-plugin) (which works pretty fine at my side)

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,27 +42,29 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
...
-                "\n" +
-                "apply(plugin = \"kotlinx-atomicfu\")\n" +
-                "```\n" +
-                "For further information, refer to the project's README: https://github.com/Kotlin/kotlinx-atomicfu/blob/master/README.md#apply-plugin"}
+        val classpath = project.buildscript.configurations.getByName("classpath")
+        val pluginVersion = classpath.resolvedConfiguration.resolvedArtifacts
+            .map { artifact -> artifact.moduleVersion.id }
+            .firstOrNull { id -> "org.jetbrains.kotlinx" == id.group && "atomicfu-gradle-plugin" == id.name }?.version
+            ?: error(
+                "We were unable to find the version of `kotlinx-atomicfu` plugin applied to the project.\n" +
```
mvicsokolova, line 50: I guess, we could get rid of this error message with this approach 🤔

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,27 +42,29 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
...
-                "    }\n" +
-                "}\n" +
-                "\n" +
-                "apply(plugin = \"kotlinx-atomicfu\")\n" +
-                "```\n" +
-                "For further information, refer to the project's README: https://github.com/Kotlin/kotlinx-atomicfu/blob/master/README.md#apply-plugin"}
+        val classpath = project.buildscript.configurations.getByName("classpath")
+        val pluginVersion = classpath.resolvedConfiguration.resolvedArtifacts
+            .map { artifact -> artifact.moduleVersion.id }
+            .firstOrNull { id -> "org.jetbrains.kotlinx" == id.group && "atomicfu-gradle-plugin" == id.name }?.version
```
mvicsokolova, line 48: And it may be `singleOrNull`

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,15 +42,40 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
 open class AtomicFUGradlePlugin : Plugin<Project> {
     override fun apply(project: Project) = project.run {
         checkCompatibility()
-        val pluginVersion = rootProject.buildscript.configurations.findByName("classpath")
-            ?.allDependencies?.find { it.name == "atomicfu-gradle-plugin" }?.version
+        val classpath = project.buildscript.configurations.getByName("classpath")
+        val pluginVersion = classpath.resolvedConfiguration.resolvedArtifacts
```
whyoleg, line 46: My bad, looks like this could not work in all cases.
From [last comment in a thread](https://discuss.gradle.org/t/how-can-a-custom-gradle-plugin-determine-its-own-version/36761/4):
> However, it fails for a multi-module gradle project where the plugin is added in the root projects as apply false, then applied without a version in the subproject.

Sorry, that I've sent a link to the specific message instead of just a thread :)

I was more thinking of the approach with a `.properties` file with version in it (mentioned in [second comment](https://discuss.gradle.org/t/how-can-a-custom-gradle-plugin-determine-its-own-version/36761/3)) if we don't want to bring external plugin for such simple thing (which is fine).

Also, I've found more [recent thread](https://discuss.gradle.org/t/how-to-get-plugin-itselfs-version/46413) (Sep 2023) where there is even better suggestion without a need for a file generation, as it's the safest and conventional way.

And BTW KGP does the same!
* build setup is [here](https://github.com/JetBrains/kotlin/blob/a5ef668e3c4d15778ea3d072221c6e379b5fb4c2/libraries/tools/kotlin-gradle-plugin/build.gradle.kts#L134-L146), though most likely the task name here will be just `processResources`
* template `project.properties` file is [here](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/common/resources/project.properties)
* function to read version is [here](https://github.com/JetBrains/kotlin/blob/a5ef668e3c4d15778ea3d072221c6e379b5fb4c2/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/utils/resourceUtils.kt#L17-L24)
* example of usage is [here](https://github.com/JetBrains/kotlin/blob/a5ef668e3c4d15778ea3d072221c6e379b5fb4c2/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPluginWrapper.kt#L386-L403)
* And additionally, just may be it will also help, here is the [commit](https://github.com/JetBrains/kotlin/commit/acf8a0454fa99e402c957ce675df7338714b15f6) in KGP which adds additional property to this file, so it touches all affected places mentioned earlier (though, it's from 2019, so some code can be moved already)

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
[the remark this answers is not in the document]
whyoleg: sorry, the comment should be left here: https://github.com/Kotlin/kotlinx-atomicfu/pull/405#discussion_r1506167680

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
mvicsokolova: Thank you for the great navigation 🔥!) Seems, that this approach worked for me.
I've pushed a commit with `atomicfu.properties` file generation.

whyoleg (approved):
Nice!

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,15 +42,24 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
...
         applyAtomicfuCompilerPlugin()
         configureDependencies()
         configureTasks()
     }
 }
 
+private fun loadPropertyFromResources(propFileName: String, property: String): String {
+    val props = Properties()
+    val inputStream = AtomicFUGradlePlugin::class.java.classLoader!!.getResourceAsStream(propFileName)
+        ?: throw FileNotFoundException("'$propFileName' file was not found in the classpath")
```
fzhinkin, line 58: Suppose I faced this exception in a project that uses AFU, what should I do? File an issue or debug my configuration? It's worth explicitly printing the next step a user has to make.

mvicsokolova: I guess, the file may not be found with older versions of atomicfu-gradle-plugin on the classpath. So, smth like this may help 👀:

"You are applying `kotlinx-atomicfu` plugin of version `0.23.3` or newer, yet we were unable to determine the specific version of the plugin. 
Starting from version `0.23.3` of `kotlinx-atomicfu`, the plugin version is extracted from the `atomicfu.properties` file, which resides within the `atomicfu-gradle-plugin-{version}.jar`. However, this file couldn't be found. 
Please ensure that there are no `atomicfu-gradle-plugin-{version}.jar` with version older than `0.23.3` present on the classpath."

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -42,15 +42,26 @@ private const val MIN_SUPPORTED_KGP_VERSION = "1.7.0"
...
         configureTasks()
     }
 }
 
+private fun loadPropertyFromResources(propFileName: String, property: String): String {
+    val props = Properties()
+    val inputStream = AtomicFUGradlePlugin::class.java.classLoader!!.getResourceAsStream(propFileName)
+        ?: throw FileNotFoundException("You are applying `kotlinx-atomicfu` plugin of version 0.23.3 or newer, yet we were unable to determine the specific version of the plugin.\". \n" +
+                "Starting from version 0.23.3 of `kotlinx-atomicfu`, the plugin version is extracted from the `atomicfu.properties` file, which resides within the atomicfu-gradle-plugin-{version}.jar. \n" +
+                "However, this file couldn't be found. Please ensure that there are no atomicfu-gradle-plugin-{version}.jar with version older than 0.23.3 present on the classpath.\n" )
```
fzhinkin, line 60: But if there is such a jar file, this code won't invoked, will it?
Maybe we can suggest filing an issue instead?

fzhinkin (approved):

Issue #399, opened 2024-01-31
ClassNotFoundException AtomicfuKotlinGradleSubplugin$AtomicfuKotlinGradleExtension


lppedd:
AtomicFU 0.23.2 with Kotlin 1.9.22.
Happens when importing changes in the IDE, or when attempting to execute any Gradle task.

```
java.lang.NoClassDefFoundError: org/jetbrains/kotlinx/atomicfu/gradle/AtomicfuKotlinGradleSubplugin$AtomicfuKotlinGradleExtension
	at kotlinx.atomicfu.plugin.gradle.AtomicFUGradlePluginKt.applyAtomicfuCompilerPlugin(AtomicFUGradlePlugin.kt:79)
	at kotlinx.atomicfu.plugin.gradle.AtomicFUGradlePluginKt.access$applyAtomicfuCompilerPlugin(AtomicFUGradlePlugin.kt:1)
	at kotlinx.atomicfu.plugin.gradle.AtomicFUGradlePlugin.apply(AtomicFUGradlePlugin.kt:48)
	at kotlinx.atomicfu.plugin.gradle.AtomicFUGradlePlugin.apply(AtomicFUGradlePlugin.kt:42)
	...
	at Program.execute(Unknown Source)
	...
Caused by: java.lang.ClassNotFoundException: org.jetbrains.kotlinx.atomicfu.gradle.AtomicfuKotlinGradleSubplugin$AtomicfuKotlinGradleExtension
	... 214 more
```

lppedd:
@fzhinkin sure, will come up with a repro soon. Seems like I'm the only one experiencing it, so it would be cool to understand where I'm breaking it.

willflier:
I also encountered this issue when upgrading from 0.32.1 to 0.32.2.

lppedd:
Are you using some third party Gradle plugin?  
Just to match what I have.

m-sasha:
Also experiencing this in [Compose Multiplatform](https://github.com/JetBrains/compose-multiplatform-core).

lppedd:
The workaround seems to be working. Thanks @m-sasha

fzhinkin:
@m-sasha thanks! 

The reproducer:
```
git clone https://github.com/JetBrains/compose-multiplatform-core.git --depth=1 -b v1.6.0-dev1405
./jbdeps/android-sdk/downloadAndroidSdk
[...]

PR #406, opened 2024-02-29
WIP: Compiler plugin is not resolved

mvicsokolova:
This is a reproducer for the problem #399 

According to our last solution (https://github.com/Kotlin/kotlinx-atomicfu/pull/386), `atomicfu-gradle-plugin` provides a transitive atomicfu compiler plugin dependency and relies on the gradle version resolution. So, we provide compiler plugin with the lowest version and expect it to be resolved into the highest version present on the user's classpath.
```
runtimeOnly "org.jetbrains.kotlin:atomicfu:1.6.21"
```

But this does not work for the project configuration presented in this PR. This is a minimized version of an MPP project created by a MPP Wizard, which uses `libs.versions.toml` file as a version catalog. 

How to reproduce the failure: 
```
cd integration-testing/examples/compiler-plugin-resolution-bug
gradle clean build
```

The build will fail with the error: 

```
* Where:
Build file '/Users/Maria.Sokolova/IdeaProjects/kotlinx-atomicfu-new/integration-testing/examples/compiler-plugin-resolution-bug/shared/build.gradle.kts' line: 16

* What went wrong:
org/jetbrains/kotlinx/atomicfu/gradle/AtomicfuKotlinGradleSubplugin$AtomicfuKotlinGradleExtension
> org.jetbrains.kotlinx.atomicfu.gradle.AtomicfuKotlinGradleSubplugin$AtomicfuKotlinGradleExtension
```

NOTE: if contents of compiler-plugin-resolution-bug/build.gradle.kts are removed, the build completes successfully.


mvicsokolova:
The last commit suggests an alternative solution to the problem of 2 atomicfu gradle plugins (#370):
the logic of the gradle plugin from the Kotlin repo (`org.jetbrains.kotlinx.atomicfu.gradle.AtomicfuKotlinGradleSubplugin`) that applies atomicfu compiler plugin transformations may be moved to the library starting from Kotlin 1.9.0.
It's possible because the sources of the compiler plugin are being published since Kotlin 1.9.0: [kotlin-atomicfu-compiler-plugin-embeddable](https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-atomicfu-compiler-plugin-embeddable/)

This will solve the problem of Kotlin version resolution "before" `kotlinx-atomicfu` plugin application.

The current solution works in overall, but still has several issues to be solved:
1. As the `kotlin.native.version` may be overriden by a user and differ from the KGP version, the Native part of the compiler plugin should be applied with  kotlin version equal to`kotlin.native.version`. For now, the Native part of the compiler plugin is separated from the JVM and JS parts but `kotlin.native.version` is not extracted.
2. Till this moment, we didn't pay attention to the fact that KGP version may differ from the version of Kotlin compiler. In `atomicfu-gradle-plugin` we extract KGP version. Though there is no public API for that yet (I'll attach the issue here).
3. A test should be added for the case if `kotlinx-atomicfu` is applied to the project that uses Kotlin version older than 1.9.0

mvicsokolova:
For now, I think it's reasonable to postpone solution of the problems 1 and 2 mentioned above until we have public API to retrieve Kotlin compiler version and Kotlin native version. Here is the corresponding ticket: [KT-66384](https://youtrack.jetbrains.com/issue/KT-66384/Public-API-to-get-Kotlin-compiler-version). 

I'll remove the separation of JVM/JS and Native compiler plugins and use KGP version to get the compiler plugin artifact, as it was done before.

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -89,48 +89,13 @@ private fun Project.applyAtomicfuCompilerPlugin() {
...
-            // for KGP >= 1.6.20 && KGP <= 1.7.20:
-            // compiler plugin for JS IR is applied via the property `kotlinx.atomicfu.enableIrTransformation`
-            // compiler plugin for JVM IR is not supported yet
-            if (kotlinVersion.atLeast(1, 6, 20)) {
-                if (rootProject.getBooleanProperty(ENABLE_JS_IR_TRANSFORMATION_LEGACY)) {
-                    plugins.apply(AtomicfuKotlinGradleSubplugin::class.java)
-                }
-            }
-        }
+        error("You are applying `kotlinx-atomicfu` plugin of version 0.23.3 or newer. " +
```
mvicsokolova, line 92: Previously, in case of KGP version < `1.9.0`, I asked users to add `org.jetbrains.kotlin:atomicfu` dependency manually to the buildscript. However, this caused further problems, that already were solved in higher Kotlin versions. 

E.g. the problem below would occur if you try to apply the current kotlinx-atomicfu version with Kotlin 1.8.20 and just add the `kotlin:atomicfu:1.8.20` dependency manually (_this problem was solved in `0.23.0` and Kotlin `1.9.10`_)

```
e: Could not find ".gradle/caches/modules-2/files-2.1/org.jetbrains.kotlinx/atomicfu-linuxarm64/0.23.2/7b4270560c2723dc55b1d3c8925b7b93201c437b/atomicfu.klib" in [IdeaProjects/kotlinx-atomicfu-new/integration-testing/examples/mpp-sample, .konan/klib, .konan/kotlin-native-prebuilt-macos-aarch64-1.8.20/klib/common, .konan/kotlin-native-prebuilt-macos-aarch64-1.8.20/klib/platform/linux_arm64]
```


So, I decided to write this error message about incompatibility instead.
I've checked that the sample mpp project can be built with kotlinx-atomicfu `0.22.0` and Kotlin down to `1.6.0`, though I believe these are very marginal use cases.

fzhinkin: > version 0.23.3 or newer

Can we print the exact version here?

I think it should be enough to:
- mention that Kotlin version should be upgraded
- add a document (like readme) describing an alternative solution for those who can't upgrade the Kotlin for some reason

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicfuKotlinCompilerPluginInternal.kt
```diff
@@ -0,0 +1,41 @@
+package kotlinx.atomicfu.plugin.gradle
```
fzhinkin, line 1: Just in case, the copyright header is not missing, but was not added intentionally because now we can, right?

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -83,28 +83,18 @@ private fun Project.checkCompatibility() {
...
-            if (rootProject.getBooleanProperty(ENABLE_JS_IR_TRANSFORMATION_LEGACY)) {
-                plugins.apply(AtomicfuKotlinGradleSubplugin::class.java)
-            }
-        }
+        error("You are applying `kotlinx-atomicfu` plugin of version 0.23.3 or newer. " +
+                "However, this version of the plugin is only compatible with Kotlin versions newer than 1.9.0.\n" +
+                "If you wish to use this version of the plugin, please upgrade your Kotlin version to 1.9.0 or newer.\n" +
+                "The alternative solution is to downgrade `kotlinx-atomicfu` plugin version to 0.22.0.\n" +
+                "Please note, that using the latest version of the plugin and upgrading the Kotlin version is a more preferable option.\n\n" +
+                "If you encounter any problems, please submit the issue: https://github.com/Kotlin/kotlinx-atomicfu/")
```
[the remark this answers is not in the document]
fzhinkin: And it's worth updating the link in the `loadPropertyFromResources` too

fzhinkin (changes requested):

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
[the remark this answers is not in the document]
mvicsokolova: Fixed)

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
mvicsokolova: I've added a note about downgrading atomicfu version as an alternative solution in the README. And in the message left just the recommendation to upgrade the Kotlin version.

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicfuKotlinCompilerPluginInternal.kt
mvicsokolova: Yes, that was intentional.  And we actually can, right?)

fzhinkin: We can, but "it would be nice to include it, but also not strictly necessary"

README.md
```diff
@@ -34,11 +34,15 @@
 
 ## Requirements
 
-Starting from version `0.23.2` of the library your project is required to use:
+Starting from version `0.23.3` of the library your project is required to use:
```
fzhinkin, line 37: I would consider changing the way we apply the plugin a major change, thus it might make sense to bump up the version to `0.24.0.`. @mvicsokolova WDYT?

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicfuKotlinCompilerPluginInternal.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.gradle.api.provider.Provider
+import org.jetbrains.kotlin.gradle.plugin.*
+
+/**
+ * This Gradle plugin applies compiler transformations to the project, it was copied from the kotlin repo (org.jetbrains.kotlinx.atomicfu.gradle.AtomicfuKotlinGradleSubplugin).
+ * 
+ * As the sources of the compiler plugin are published as `org.jetbrains.kotlin.kotlin-atomicfu-compiler-plugin-embeddable` starting from Kotlin 1.9.0,
+ * the Gradle plugin can access this artifact from the library and apply the transformations.
+ * 
+ * NOTE: The version of KGP may differ from the version of Kotlin compiler, and kotlin.native.version may override the version of Kotlin native compiler.
```
fzhinkin, line 12: It's worth mentioning the constraint in README.

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicfuKotlinCompilerPluginInternal.kt
```diff
@@ -0,0 +1,41 @@
...
+ * So, the right behavior for the Gradle plugin would be to obtain compiler versions and apply compiler transformations separately to JVM/JS and Native targets.
+ * This was postponed as a separate task (#408).
+ */
+internal class AtomicfuKotlinCompilerPluginInternal : KotlinCompilerPluginSupportPlugin {
+    
+    companion object {
+        const val ATOMICFU_ARTIFACT_NAME = "kotlin-atomicfu-compiler-plugin-embeddable"
+    }
+
```
fzhinkin, line 21:
```suggestion

```

Please, cleanup file's formatting.

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -81,30 +79,19 @@ private fun Project.checkCompatibility() {
...
     } else {
-        // for KGP >= 1.6.20 && KGP <= 1.7.20:
-        // compiler plugin for JS IR is applied via the property `kotlinx.atomicfu.enableIrTransformation`
-        // compiler plugin for JVM IR is not supported yet
-        if (kotlinVersion.atLeast(1, 6, 20)) {
-            if (rootProject.getBooleanProperty(ENABLE_JS_IR_TRANSFORMATION_LEGACY)) {
-                plugins.apply(AtomicfuKotlinGradleSubplugin::class.java)
-            }
-        }
+        error("You are applying `kotlinx-atomicfu` plugin of version $afuPluginVersion. " +
```
fzhinkin, line 90: We already have `Project.checkCompatibility` checking that Kotlin is at least `1.7.0`. Probably this error should be moved there.

README.md
```diff
@@ -34,11 +34,15 @@
...
 
 * Gradle `7.0` or newer
```
fzhinkin, line 39: Below in this file: `Gradle configuration is supported for all platforms, minimal version is Gradle 6.8.`

README.md
mvicsokolova: Right, it'd be better to release `0.24.0`

README.md
```diff
@@ -34,11 +34,15 @@
...
 
-* Kotlin `1.7.0` or newer
+* Kotlin `1.9.0` or newer
```
fzhinkin, line 41: `## Atomicfu compiler plugin` sections seems to be obsolete as it describes how the plugin should be configured for older versions.

As there's a suggestion to use an older atomicfu version with older Kotlin version, it does make sense to move everything related to older versions to a dedicated page (or leave it here, but stress that it only applies to older atomicfu versions).

README.md
mvicsokolova: Thanks, didn't notice that. I'd remove this line completely.

atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicfuKotlinCompilerPluginInternal.kt
mvicsokolova: Added **Note on Kotlin version** in the README

README.md
mvicsokolova: I've fixed this section, added a note about Kotlin version requirements for the current atomicfu version and moved old configurations to the hidden section.

PR #418, opened 2024-04-18
Moved all the dependencies version to version catalog

dkrasnoff:
In this PR we moved all the dependencies version in [libs.versions.toml](https://docs.gradle.org/8.4/userguide/platforms.html#sub:conventional-dependencies-toml). Such approach have several advantages:

- Gradle generates type-safe accessors for each dependency
- versions catalog is visible for all projects, so it is a single place to declare all versions

### DIFF
--- a/atomicfu-gradle-plugin/build.gradle
+++ b/atomicfu-gradle-plugin/build.gradle
@@ -19,16 +19,16 @@ dependencies {
     }
 
     compileOnly gradleApi()
-    compileOnly 'org.jetbrains.kotlin:kotlin-stdlib'
-    compileOnly "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+    compileOnly libs.kotlin.stdlib
+    compileOnly libs.kotlin.gradlePlugin
     // Atomicfu compiler plugin dependency will be loaded to kotlinCompilerPluginClasspath
     // Atomicfu plugin will only be applied if the flag is set kotlinx.atomicfu.enableJsIrTransformation=true
-    compileOnly "org.jetbrains.kotlin:atomicfu:$kotlin_version"
+    compileOnly libs.kotlin.atomicfu
 
     testImplementation gradleTestKit()
-    testImplementation 'org.jetbrains.kotlin:kotlin-test'
-    testImplementation 'org.jetbrains.kotlin:kotlin-test-junit'
-    testImplementation 'junit:junit:4.12'
+    testImplementation libs.kotlin.test
+    testImplementation libs.kotlin.testJunit
+    testImplementation libs.junit.junit
 }
 
 evaluationDependsOn(':atomicfu')
--- a/buildSrc/settings.gradle.kts
+++ b/buildSrc/settings.gradle.kts
@@ -1,0 +1,32 @@
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+
+            val kotlinVersion = providers.gradleProperty("kotlin_version").orNull
+            if (kotlinVersion != null) {
+                version("kotlin", kotlinVersion)
+            }
+
+            val overridingKotlinVersion =  providers.gradleProperty("community.project.kotlin.version").orNull
+            if (overridingKotlinVersion != null) {
+                logger.info("An overriding Kotlin version of $overridingKotlinVersion was found for buildSrc")
+                version("kotlin", overridingKotlinVersion)
+            }
+
+            /*
+            * This property group is used to build kotlinx.atomicfu against Kotlin compiler snapshots.
+            * When build_snapshot_train is set to true, kotlin_version property is overridden with kotlin_snapshot_version.
+            * Additionally, mavenLocal and Sonatype snapshots are added to repository list
+            * (the former is required for AFU and public, the latter is required for compiler snapshots).
+            * DO NOT change the name of these properties without adapting kotlinx.train build chain.
+            */
+            val buildSnapshotTrainGradleProperty = providers.gradleProperty("build_snapshot_train").orNull
+            if (buildSnapshotTrainGradleProperty != null && buildSnapshotTrainGradleProperty != "") {
+                val kotlinVersion = providers.gradleProperty("kotlin_snapshot_version").orNull
+                    ?: throw IllegalArgumentException("'kotlin_snapshot_version' should be defined when building with a snapshot compiler")
+                version("kotlin", kotlinVersion)
+            }
+        }
+    }
+}
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,0 +1,50 @@
+[versions]
+kotlin = "1.9.21"
+kotlinx-metadata = "0.7.0"
+asm = "9.6"
+slf4j = "1.8.0-alpha2"
+junit = "4.12"
+maven = "3.5.3"
+maven-pluginTools = "3.5"
+node-gradle = "3.1.1"
+node = "8.11.1"
+npm = "5.7.1"
+rhino = "1.7.10"
+
+[libraries]
+
+# Kotlin dependencies
+kotlin-gradlePlugin = { group = "org.jetbrains.kotlin", name = "kotlin-gradle-plugin", version.ref = "kotlin" }
+kotlin-atomicfu = { group = "org.jetbrains.kotlin", name = "atomicfu", version.ref = "kotlin" }
+kotlin-stdlib = { group = "org.jetbrains.kotlin", name = "kotlin-stdlib", version.ref = "kotlin" }
+kotlin-stdlibJdk8 = { group = "org.jetbrains.kotlin", name = "kotlin-stdlib-jdk8", version.ref = "kotlin" }
+kotlin-test = { group = "org.jetbrains.kotlin", name = "kotlin-test", version.ref = "kotlin" }
+kotlin-testJunit = { group = "org.jetbrains.kotlin", name = "kotlin-test-junit", version.ref = "kotlin" }
+kotlin-scriptRuntime = { group = "org.jetbrains.kotlin", name = "kotlin-script-runtime", version.ref = "kotlin" }
+kotlinx-metadataJvm = { group = "org.jetbrains.kotlinx", name = "kotlinx-metadata-jvm", version.ref = "kotlinx-metadata" }
+
+# ASM dependencies
+asm = { group = "org.ow2.asm", name = "asm", version.ref = "asm" }
+asm-commons = { group = "org.ow2.asm", name = "asm-commons", version.ref = "asm" }
+asm-tree = { group = "org.ow2.asm", name = "asm-tree", version.ref = "asm" }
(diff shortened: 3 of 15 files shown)


settings.gradle.kts
```diff
@@ -0,0 +1,34 @@
...
+include("atomicfu-maven-plugin")
+
+include("atomicfu-native") // this is a kludge so that existing YouTrack config works, todo: remove
+include("integration-testing")
+
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("libs") {
+
+            val overridingKotlinVersion =  providers.gradleProperty("community.project.kotlin.version").orNull
```
dkrasnoff, line 13: after moving this property here, I have a doubled dependency declaration for `kotlin-script-runtime` in `integration-testing` artifact. It looks like:
```
"dependencies": [
        {
          "group": "org.jetbrains.kotlin",
          "module": "kotlin-stdlib-jdk8",
          "version": {
            "requires": "1.9.21"
          }
        },
        {
          "group": "org.jetbrains.kotlin",
          "module": "kotlin-script-runtime'",
          "version": {
            "requires": "1.9.21"
          }
        },
        {
          "group": "org.jetbrains.kotlin",
          "module": "kotlin-script-runtime",
          "version": {
            "requires": "1.9.21"
          }
        }
      ],
```
I'm not sure in reason, but don't see a big problem here

dkrasnoff: We found a mistake in toml file

dkrasnoff: Oh, with fix we now have identical duplicated blocks

buildSrc/settings.gradle.kts
```diff
@@ -0,0 +1,32 @@
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+
+            val kotlinVersion = providers.gradleProperty("kotlin_version").orNull
```
dkrasnoff, line 6: I don't like that we have 3 ways of overriding Kotliin version. So I'm planning to discuss it with the author, and maybe after some changes in our infrastructure build scripts, I'll leave only one.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,49 @@
...
+npm = "5.7.1"
+rhino = "1.7.10"
+
+[libraries]
+
+# Kotlin dependencies
+kotlin-gradlePlugin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
+kotlin-atomicfu = { module = "org.jetbrains.kotlin:atomicfu", version.ref = "kotlin" }
+kotlin-stdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
+kotlin-stdlibJdk8 = { group = "org.jetbrains.kotlin", name = "kotlin-stdlib-jdk8", version.ref = "kotlin" }
```
adam-enko, line 19: Are the Kotlin stdlib dependencies necessary? They should be added automatically by KGP (since it doesn't define `kotlin.stdlib.default.dependency=false`), so they don't need to be added here, and can be removed from the build scripts?

Maybe this PR is just accurately converting the existing dependencies, and the Kotlin stdlib dependencies can be revisited in later PR?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,49 @@
...
+slf4j-api = { group = "org.slf4j", name = "slf4j-api", version.ref = "slf4j" }
+slf4j-simple = { group = "org.slf4j", name = "slf4j-simple", version.ref = "slf4j" }
+
+# Testing depedencies
+junit-junit = { group = "junit", name = "junit", version.ref = "junit" }
+
+# Maven dependencies
+maven-core = { group = "org.apache.maven", name = "maven-core", version.ref = "maven" }
+maven-pluginApi = { group = "org.apache.maven", name = "maven-plugin-api", version.ref = "maven" }
+maven-pluginAnnotations = { group = "org.apache.maven.plugin-tools", name = "maven-plugin-annotations", version = "3.5" }
```
adam-enko, line 41: Nitpick:

I think it's a good idea to have all versions defined under `[versions]`. It helps with consistency, and can help minimise changes if, in the future, a version does end up being re-used.

So, I'd suggest changing `version = "3.5"` to `version.ref = "maven-pluginTools"`, and adding a version.

(Just for reference, Dokka has the same `maven-plugin-annotations` dependency: https://github.com/Kotlin/dokka/blob/4f5a650d6ae8f4f94c34ae4bedf72e7c660aecbe/gradle/libs.versions.toml#L128)

gradle/libs.versions.toml
```diff
@@ -0,0 +1,49 @@
...
+
+# Kotlin dependencies
+kotlin-gradlePlugin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
+kotlin-atomicfu = { module = "org.jetbrains.kotlin:atomicfu", version.ref = "kotlin" }
+kotlin-stdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
+kotlin-stdlibJdk8 = { group = "org.jetbrains.kotlin", name = "kotlin-stdlib-jdk8", version.ref = "kotlin" }
+kotlin-test = { group = "org.jetbrains.kotlin", name = "kotlin-test", version.ref = "kotlin" }
+kotlin-testJunit = { group = "org.jetbrains.kotlin", name = "kotlin-test-junit", version.ref = "kotlin" }
+kotlin-scriptRuntime = { group = "org.jetbrains.kotlin", name = "kotlin-script-runtime", version.ref = "kotlin" }
+kotlinx-metadataJvm = { group = "org.jetbrains.kotlinx", name = "kotlinx-metadata-jvm", version.ref = "kotlinx-metadata" }
```
adam-enko, line 23: Nitpick:

It'd be nicer to use `module` consistently (it's more compact and more closely matches how dependencies are usually defined in build scripts), rather than a mix of `module` and `group`/`name`.

For example:

```suggestion
# Kotlin dependencies
kotlin-gradlePlugin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
kotlin-atomicfu = { module = "org.jetbrains.kotlin:atomicfu", version.ref = "kotlin" }
kotlin-stdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
kotlin-stdlibJdk8 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", version.ref = "kotlin" }
kotlin-test = { module = "org.jetbrains.kotlin:kotlin-test", version.ref = "kotlin" }
kotlin-testJunit = { module = "org.jetbrains.kotlin:kotlin-test-junit", version.ref = "kotlin" }
kotlin-scriptRuntime = { module = "org.jetbrains.kotlin:kotlin-script-runtime", version.ref = "kotlin" }
kotlinx-metadataJvm = { module = "org.jetbrains.kotlinx:kotlinx-metadata-jvm", version.ref = "kotlinx-metadata" }
```

adam-enko:
I see some dependencies in `atomicfu/build.gradle` that haven't been updated to use the version catalog - is this intentional, or should they be converted too?

atomicfu-gradle-plugin/build.gradle
```diff
@@ -20,15 +20,15 @@ dependencies {
 
     compileOnly gradleApi()
     compileOnly 'org.jetbrains.kotlin:kotlin-stdlib'
-    compileOnly "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+    compileOnly libs.kotlin.gradlePlugin
     // Atomicfu compiler plugin dependency will be loaded to kotlinCompilerPluginClasspath
     // Atomicfu plugin will only be applied if the flag is set kotlinx.atomicfu.enableJsIrTransformation=true
-    compileOnly "org.jetbrains.kotlin:atomicfu:$kotlin_version"
+    compileOnly "org.jetbrains.kotlin:atomicfu:${libs.versions.kotlin}"
```
adam-enko, line 26: I'm not familiar with the Groovy DSL, will this get the actual value, or instead call `Provide<String>.toString()`, which returns something like `"Provider(...)"` instead of `1.9.23`?

In Kotlin you need to `.get()` the value, e.g.:

```suggestion
    compileOnly "org.jetbrains.kotlin:atomicfu:${libs.versions.kotlin.get()}"
```

atomicfu-gradle-plugin/build.gradle
```diff
@@ -20,15 +20,15 @@ dependencies {
 
     compileOnly gradleApi()
     compileOnly 'org.jetbrains.kotlin:kotlin-stdlib'
```
adam-enko, line 22: Should this dependency be updated too, or removed (because kotlin-stdlib should be added automatically)?

adam-enko (changes requested):

buildSrc/settings.gradle.kts
adam-enko: Good idea - can you make an issue to keep track, and add a `TODO <link to issue>` in the code?

It'd also be a good idea to de-duplicate the version override in the actual project and in buildSrc.

dkrasnoff: added https://github.com/Kotlin/kotlinx-atomicfu/issues/421

gradle/libs.versions.toml
dkrasnoff: fixed

gradle/libs.versions.toml
dkrasnoff: fixed

atomicfu-gradle-plugin/build.gradle
dkrasnoff: fixed

atomicfu-gradle-plugin/build.gradle
dkrasnoff: fixed

gradle/libs.versions.toml
adam-enko: Nice!

(For context, Dmitrii and I had a call and we agreed to use group/name consistently for all dependencies, and not use module.)

adam-enko (approved):
Good job!

mvicsokolova (approved):

adam-enko:
@dkrasnoff the tests are failing on TeamCity - is this expected? https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxAtomicfu_BuildAggregatedLinuxWindows/4585035?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true

dkrasnoff:
I see that it is pretty frequent fail for our [win builds](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxAtomicfu_BuildWindows?mode=builds#all-projects), it seems like flaky fails, Fortunately, on my machine I could reproduce it.
And I found that [YT issue](https://youtrack.jetbrains.com/issue/KT-60572).
So, I made an additional "Fix KT-60572 with Synchronized.kt".
@adam-enko wdyt?

PR #420, opened 2024-04-19
Run integration tests with Kotlin version used to build the library

### LINKED ISSUE #419: Pass Kotlin version as a parameter to the projects from integration-testing
Currently, `integration-testing` projects are built with Kotlin version hardcoded in `gradle.properties` file of each project. It would be better to use the Kotlin version passed as a parameter to the test the project build: by default it would be the same Kotlin version as used in the library, and it can be overriden by TC, e.g. for the Aggregate build.

This behaviour also causes compatibility problems while running an Aggregate build and prevents to test these changes: [KT-49279](https://youtrack.jetbrains.com/issue/KT-49279).  After updating LLVM, cinterop produced klibs can't be consumed by compiler with version `1.9.21`.

mvicsokolova:
Fixes #419

The main goal of this PR is to test builds of projects from `integration-testing` against the Kotlin version used to build the library. E.g. in Aggregate build test projects should be built with Kotlin from master branch. Otherwise with the same Kotlin version that is set in the root `gradle.properties` file.

Previously Kotlin version for all integration tests was hardcoded in their local `gradle.properties` files.

### DIFF
--- a/integration-testing/build.gradle.kts
+++ b/integration-testing/build.gradle.kts
@@ -65,7 +65,7 @@ val functionalTest by tasks.registering(Test::class) {
     classpath = sourceSets["functionalTest"].runtimeClasspath
 
     // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
-    systemProperties["kotlin.version.integration"] = libs.versions.kotlin
+    systemProperties["kotlin.version.integration"] = providers.gradleProperty("kotlin_version").orNull
     // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
     systemProperties["kotlin.native.version.integration"] = providers.gradleProperty("kotlin.native.version").orNull
     // the current atomicfu version set in the root gradle.properties
--- a/integration-testing/build.gradle.kts
+++ b/integration-testing/build.gradle.kts
@@ -24,9 +24,6 @@ kotlin {
     jvmToolchain(11)
 }
 
-val kotlin_version = providers.gradleProperty(libs.versions.kotlin).orNull
-val atomicfu_snapshot_version = providers.gradleProperty("version").orNull
-
 sourceSets {
     create("mavenTest") {
         compileClasspath += files(sourceSets.main.get().output, configurations.testRuntimeClasspath)
@@ -68,7 +65,7 @@ val functionalTest by tasks.registering(Test::class) {
     classpath = sourceSets["functionalTest"].runtimeClasspath
 
     // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
-    systemProperties["kotlin.version.integration"] = providers.gradleProperty("kotlin_version").orNull
+    systemProperties["kotlin.version.integration"] = libs.versions.kotlin
     // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
     systemProperties["kotlin.native.version.integration"] = providers.gradleProperty("kotlin.native.version").orNull
     // the current atomicfu version set in the root gradle.properties
--- a/integration-testing/build.gradle.kts
+++ b/integration-testing/build.gradle.kts
@@ -72,7 +72,7 @@ val functionalTest by tasks.registering(Test::class) {
     // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
     systemProperties["kotlin.native.version.integration"] = providers.gradleProperty("kotlin.native.version").orNull
     // the current atomicfu version set in the root gradle.properties
-    systemProperties["atomicfu.version.integration"] = providers.gradleProperty("version").orNull
+    systemProperties["atomicfu.snapshot.version.integration"] = providers.gradleProperty("version").orNull
     // the directory (on TC agent) where Kotlin artifacts were published during the Aggregate build
     systemProperties["kotlin.artifacts.repository.integration"] = providers.gradleProperty("kotlin_repo_url").orNull
 
--- a/integration-testing/build.gradle.kts
+++ b/integration-testing/build.gradle.kts
@@ -1,3 +1,7 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
 import org.jetbrains.kotlin.gradle.utils.NativeCompilerDownloader
 
 plugins {
@@ -63,8 +67,14 @@ val functionalTest by tasks.registering(Test::class) {
     testClassesDirs = sourceSets["functionalTest"].output.classesDirs
     classpath = sourceSets["functionalTest"].runtimeClasspath
 
-    systemProperties["kotlinVersion"] = kotlin_version
-    systemProperties["atomicfuVersion"] = atomicfu_snapshot_version
+    // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
+    systemProperties["kotlin.version.integration"] = providers.gradleProperty("kotlin_version").orNull
+    // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
+    systemProperties["kotlin.native.version.integration"] = providers.gradleProperty("kotlin.native.version").orNull
+    // the current atomicfu version set in the root gradle.properties
+    systemProperties["atomicfu.version.integration"] = providers.gradleProperty("version").orNull
+    // the directory (on TC agent) where Kotlin artifacts were published during the Aggregate build
+    systemProperties["kotlin.artifacts.repository.integration"] = providers.gradleProperty("kotlin_repo_url").orNull
 
     dependsOn(":atomicfu-gradle-plugin:publishToMavenLocal")
     // atomicfu-transformer and atomicfu artifacts should also be published as it's required by atomicfu-gradle-plugin.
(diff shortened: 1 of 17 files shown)


mvicsokolova:
@shanshin Could you please take a look at this PR? 🙏🏻

integration-testing/build.gradle.kts
```diff
@@ -63,9 +60,11 @@ val mavenTest by tasks.registering(Test::class) {
 val functionalTest by tasks.registering(Test::class) {
     testClassesDirs = sourceSets["functionalTest"].output.classesDirs
     classpath = sourceSets["functionalTest"].runtimeClasspath
-    
-    systemProperties["kotlinVersion"] = kotlin_version
-    systemProperties["atomicfuVersion"] = atomicfu_snapshot_version
+
+    // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
+    systemProperties["kotlin_version_for_integration_tests"] = providers.gradleProperty("kotlin_version").orNull
```
mvicsokolova, line 65: Is this the right way to extract the `kotlin_version` used to build the library, including the case when the version was overriden by the TC config? 👀

mvicsokolova:
The build against Kotlin master fails, because isolated gradle projects from `integration-testing` cannot resolve the Kotlin version used in the build. Kotlin artifacts used in the Aggregate build are not published to space repo, and are passed as parameters of TC config, e.g.:

```
-Pkotlin_repo_url=${kotlinArtifactsRepositoryUrlParamRef.ref}
-Pkotlin_version=${BuildNumber.deployVersion.ref(build)} 
-Pkotlin.native.version=${BuildNumber.deployVersion.ref(build)} 
```

I'm trying to fix that by passing `project.findProperty("kotlin.artifacts.repository")` as a `kotlin_url_repo` parameter

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
```diff
@@ -31,28 +31,28 @@ class MppProjectTest {
     @Test
     fun testMppWithEnabledJsIrTransformation() {
         mppSample.enableJsIrTransformation = true
-        assertTrue(mppSample.cleanAndBuild().isSuccessful)
```
mvicsokolova, line 34: These are just minor changes of the error output format. I could've considered adding them separately tho.

integration-testing/build.gradle.kts
```diff
@@ -63,8 +67,14 @@ val functionalTest by tasks.registering(Test::class) {
     testClassesDirs = sourceSets["functionalTest"].output.classesDirs
     classpath = sourceSets["functionalTest"].runtimeClasspath
 
-    systemProperties["kotlinVersion"] = kotlin_version
-    systemProperties["atomicfuVersion"] = atomicfu_snapshot_version
+    // the kotlin version used to build the library, which is set in root gradle.properties or overriden by the TC config
+    systemProperties["kotlin.version.integration"] = providers.gradleProperty("kotlin_version").orNull
```
dkrasnoff, line 71: Wdyt of using `libs.versions.kotlin` instead of `providers.gradleProperty("kotlin_version").orNull` in such a way we will bind all ways of overridden kotlin version with `systemProperties["kotlin.version.integration"]`?

integration-testing/examples/mpp-version-catalog/shared/build.gradle.kts
```diff
@@ -8,6 +12,7 @@ buildscript {
 
 repositories {
     mavenCentral()
+    maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
```
dkrasnoff, line 15: Could you please explain why do we need `dev` repo here and in other examples? I hight don't recommend to use this repo, because it has temporary artifacts, which don't live long.
So if it not expected by TeamCity build to have, I would suggest to remove the usages of this repo.

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildScriptConfigurator.kt
```diff
@@ -0,0 +1,49 @@
...
+    val originLines = readLines()
+    var previousLine = ""
+
+    bufferedWriter().use { writer ->
+        originLines.forEach { line ->
+            writer.appendLine(line)
+            // in build.gradle exclude publishing { repositories {...} } block
+            val isInPublishingBlock = previousLine.trimStart().startsWith("publishing")
+            // in settings.gradle add repository only in pluginManagement {...} block
+            val isInPluginManagementBlock = line.trimStart().startsWith("pluginManagement")
```
dkrasnoff, line 37: `line.trimStart().startsWith("pluginManagement")` -> `previousLine.trimStart().startsWith("pluginManagement")` ?

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
```diff
@@ -19,9 +19,11 @@ internal class GradleBuild(val projectName: String, val targetDir: File) {
...
             add("-P$ENABLE_JS_IR_TRANSFORMATION=$enableJsIrTransformation")
             add("-P$ENABLE_NATIVE_IR_TRANSFORMATION=$enableNativeIrTransformation")
-            localRepositoryUrl?.let {
-                add("-P$LOCAL_REPOSITORY_URL_PROPERTY=$it")
-            }
+            // Pass kotlin_version and kotlin.native.version parameters used for the library build to the builds of integration tests.
+            // These versions may be overriden by the TC build config.
+            if (kotlinVersion.isNotEmpty()) add("-P$KOTLIN_VERSION_PARAMETER=$kotlinVersion")
+            if (kotlinNativeVersion.isNotEmpty()) add("-P$KOTLIN_NATIVE_VERSION_PARAMETER=$kotlinNativeVersion")
+            localRepositoryUrl?.let { add("-P$LOCAL_REPOSITORY_URL_PROPERTY=$it") }
```
dkrasnoff, line 26: can we get rid of `localRepositoryUrl` using `kotlinArtifactsRepo` instead of this?

integration-testing/build.gradle.kts
mvicsokolova: Fixed

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/PluginOrderBugTest.kt
```diff
@@ -14,22 +14,24 @@ class PluginOrderBugTest {
...
-     * This test ensures that kotlin-stdlib of the Kotlin version used to build kotlinx-atomicfu library is not "required" in the user's project.
+     *
+     * This test sets Kotlin version in `plugin-order-bug` project that differs from the Kotlin version set for the library.
+     * The goal is to check that kotlin-stdlib of the Kotlin version used to build kotlinx-atomicfu library is not "required" in the user's project.
      * The user project should use kotlin-stdlib version that is present in it's classpath.
      */
     @Test
     fun testTransitiveKotlinStdlibDependency() {
+        pluginOrderBugProject.extraProperties.add("-Pkotlin_version=1.9.20")
+        assertTrue(pluginOrderBugProject.getKotlinVersion() == "1.9.20")
```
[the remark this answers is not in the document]
mvicsokolova: Actually, no. This test is used to ensure that atomicfu dependency does not transitively pull `kotlin-stdlib` dependency to the user project. The user project should use `kotlin-stdlib` with `kotlin_version` used in the user project (or which one is bundled in Gradle).

integration-testing/examples/mpp-version-catalog/shared/build.gradle.kts
mvicsokolova: I guess, there were some cases, when I had to run `kotlinx-atomicfu` build against some Kotlin version published to `space.dev`, that's why I left them. Is that a problem or it's better to remove this repo anyway?

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt
mvicsokolova: There is one test (`user-project`), that is supposed to be built against the library that uses atomicfu. So `user-project` depends on `mpp-project`, that is published to `maven("build/.m2/")`. And in this case I pass `"build/.m2/" `as a `localRepositoryUrl`

integration-testing/build.gradle.kts
mvicsokolova: Mm, for some reason it didn't get resolved that way 👀

dkrasnoff: Sorry it is my mistake, I see you used `libs.versions.kotlin` as I wrote, but it will return `provider` instead of real value, so the right call is `libs.versions.kotlin.get()`.

integration-testing/examples/mpp-version-catalog/shared/build.gradle.kts
dkrasnoff: There are some points to reduce the usage of this repository:
- decrease the transfer of unnecessary traffic via the network
- decrease failures connected with unstable internet connection on agents
I recommend using the `maven.zip` artifact in our CI builds instead of this. Of course, we can create an additional issue for that.

dkrasnoff (approved):

PR #424, opened 2024-04-26
Moved repositories to setting.gradle.kts

dkrasnoff:
Put all repositories into `pluginManagement` and `dependencyResolutionManagement`. This approach creates a single place with all available repositories for all projects.
Also, for `dependencyResolutionManagement` set a `RepositoriesMode.PREFER_SETTINGS)`, which enables repositories only from setting file, but allows to configure repositories during the build (the last one will be then ignored).


settings.gradle.kts
```diff
@@ -1,12 +1,84 @@
...
+    val buildSnapshotTrainGradleProperty = providers.gradleProperty("build_snapshot_train")
+
+    @Suppress("UnstableApiUsage")
+    repositoriesMode.set(RepositoriesMode.PREFER_SETTINGS)
+
+    @Suppress("UnstableApiUsage")
+    repositories {
+
+        // TODO(Dmitrii Krasnov): maybe could be passed via additionalRepositoryProperty
+        maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
```
dkrasnoff, line 41: I requested a list of TeamCity CI builds, to check which configurations use this repository

dkrasnoff: The same about `mavenLocal()`

settings.gradle.kts
```diff
@@ -1,12 +1,84 @@
...
+                    patternLayout { artifact("v[revision]/[artifact](-v[revision]).[ext]") }
+                    metadataSources { artifact() }
+                    content { includeModule("com.yarnpkg", "yarn") }
+                }
+            }
+            filter { includeGroup("com.yarnpkg") }
+        }
+
+        // TODO(Dmitrii Krasnov): I'm not sure that we really need it
+        mavenLocal()
```
adam-enko, line 79: Yeah, it's rare that Maven Local is needed, so it should probably be removed. If a dev needs it, they can add it manually. What do you think? Should it be removed in this PR, or a later one?

settings.gradle.kts
```diff
@@ -1,12 +1,84 @@
...
+        exclusiveContent {
+            forRepository {
+                ivy("https://github.com/yarnpkg/yarn/releases/download") {
+                    patternLayout { artifact("v[revision]/[artifact](-v[revision]).[ext]") }
+                    metadataSources { artifact() }
+                    content { includeModule("com.yarnpkg", "yarn") }
+                }
+            }
+            filter { includeGroup("com.yarnpkg") }
+        }
```
adam-enko, line 76: Is there a YouTrack ticket to make manually adding the node/yarn repos unnecessary? There's https://youtrack.jetbrains.com/issue/KT-51379, but that's for the K/N repos.

settings.gradle.kts
```diff
@@ -1,12 +1,84 @@
...
 dependencyResolutionManagement {
+    val buildSnapshotTrainGradleProperty = providers.gradleProperty("build_snapshot_train")
+
+    @Suppress("UnstableApiUsage")
+    repositoriesMode.set(RepositoriesMode.PREFER_SETTINGS)
+
+    @Suppress("UnstableApiUsage")
+    repositories {
+
+        // TODO(Dmitrii Krasnov): maybe could be passed via additionalRepositoryProperty
```
adam-enko, line 40: > maybe could be passed via additionalRepositoryProperty

It's generally better to _always_ define repositories. It helps with reproducible builds, as Gradle will always fetch dependencies from the same place.

settings.gradle.kts
dkrasnoff: I see that [it was added recently](https://github.com/Kotlin/kotlinx-atomicfu/pull/392) (Jan 2024) for train builds. I'm trying to figure out is it still really necessary for train builds.
I hope that I figure out during this PR, but I believe that we can postpone removing it until the next PR.

settings.gradle.kts
dkrasnoff: Yeah, but generally this repository should not be used at all, for reasons like:
- decrease the transfer of unnecessary traffic via the network
- decrease failures connected with unstable internet connection on agents
but before I figure out how to get rid of it properly, it would be better to at least add it only when necessary.

settings.gradle.kts
dkrasnoff: I see this task: https://youtrack.jetbrains.com/issue/KT-55620/

dkrasnoff: And I've asked Ilya about current status

dkrasnoff: UPD: @adam-enko Starting from the `2.0.0` we will be able to pass `null` repository URL to plugin and then use even `RepositoriesMode.FAIL_ON_PROJECT_REPOS`, but we will need to declare these repositories anyway

dkrasnoff: UPD: there is also https://youtrack.jetbrains.com/issue/KT-34732, but it is only in backlog

settings.gradle.kts
dkrasnoff: The only configuration where we used `mavenLocal` was `train build`. I've checked whether it is still necessary and removed it.

settings.gradle.kts
dkrasnoff: Put a comment

adam-enko (approved):
👍

mvicsokolova (approved):

PR #432, opened 2024-04-29
Pass -Pversion parameter to integration tests to override atomicfu_version

mvicsokolova:


integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Environment.kt
```diff
@@ -12,6 +12,7 @@ internal const val ENABLE_NATIVE_IR_TRANSFORMATION = "kotlinx.atomicfu.enableNat
 internal const val LOCAL_REPOSITORY_URL_PROPERTY = "localRepositoryUrl"
 internal const val KOTLIN_VERSION_PARAMETER = "kotlin_version"
 internal const val KOTLIN_NATIVE_VERSION_PARAMETER = "kotlin.native.version"
+internal const val ATOMICFU_VERSION = "version"
```
dkrasnoff, line 15: `version` -> `atomicfu_version`

dkrasnoff (changes requested):

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Environment.kt
dkrasnoff: If I'm right we are talking about [this parameter](https://github.com/Kotlin/kotlinx-atomicfu/blob/9d2a3e4df18246c14aae2272012b3ecc901f16de/integration-testing/examples/mpp-sample/settings.gradle.kts#L13)

dkrasnoff: The `version` parameter you are talking about [is here](https://github.com/Kotlin/kotlinx-atomicfu/blob/bb4276afc3378a3203d223e6193474fe290b68ef/integration-testing/build.gradle.kts#L72).
So, I see the scenario like that:
1) train build uses `-Pversion=2.0.0-atomicfu..train`
2) then we [catch it here](https://github.com/Kotlin/kotlinx-atomicfu/blob/bb4276afc3378a3203d223e6193474fe290b68ef/integration-testing/build.gradle.kts#L72) and convert it to test system property `atomicfu.snapshot.version.integration`
3) then inside [BuildRunner.kt we read](https://github.com/Kotlin/kotlinx-atomicfu/blob/d7bb720684c046c9687b2723b561c977e32b90ba/integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/BuildRunner.kt#L25) test system property `atomicfu.snapshot.version.integration` and add its value to `ATOMICFU_VERSION` (which we are discussing here)
4) This `ATOMICFU_VERSION` is the property that is used in test project properties

From here, I assumed that it should be [atomicfu_version](https://github.com/Kotlin/kotlinx-atomicfu/blob/9d2a3e4df18246c14aae2272012b3ecc901f16de/integration-testing/examples/mpp-sample/settings.gradle.kts#L13) instead of `version.`
Could you please point out where I'm mistaken?

dkrasnoff: Greate that we figured it out 👍

dkrasnoff (approved):

Issue #421, opened 2024-04-22
Refactoring build scripts


dkrasnoff:
- [ ]  move plugin and dependency repositories to setting file
- [ ] change files from `gradle` to gradle convention plugins 1 by 1
- [ ]  move part of allprojects block from root build. to convention plugining
- [ ]  build.gradle -> build.gradle.kts
- [ ] update targets
- [ ] remove deprecated things like `buildDir`
- [ ]  de-duplicate the version override in the actual project and in buildSrc.
- [ ]  reduce 3 ways of overriding Kotliin version

dkrasnoff:
For test suits I've created separate issue: https://github.com/Kotlin/kotlinx-atomicfu/issues/457

PR #433, opened 2024-04-30
Extended integration tests configurations with implementation conf

dkrasnoff:
Updated integration test configurations according to [gradle docs](https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests).
It helps to IJ IDEA better resolve test source and run tests using IJ IDEA gutters

### DIFF
--- a/integration-testing/build.gradle.kts
+++ b/integration-testing/build.gradle.kts
@@ -26,13 +26,29 @@ kotlin {
 
 sourceSets {
     create("mavenTest") {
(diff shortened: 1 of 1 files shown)


adam-enko (approved):
👍 

Looks like a decent good-enough-for-now fix!

Let's add a task to https://github.com/Kotlin/kotlinx-atomicfu/issues/421 "update integration tests to use [JVM Test Suites](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html)", so the config gets fixed properly.

mvicsokolova (approved):

PR #435, opened 2024-05-10
Moved publishing.gradle to convention plugin

dkrasnoff:
Divided `publishing.gradle` into two convention plugins: `non-multiplatform.publish-conventions.gradle.kts` and `publish-conventions.gradle.kts`

Depends on

* https://github.com/Kotlin/kotlinx-atomicfu/pull/434


buildSrc/src/main/kotlin/non-multiplatform.publish-conventions.gradle.kts
```diff
@@ -0,0 +1,31 @@
+plugins {
+//     Regular java modules need 'java-library' plugin for proper publication
+    `java-library`
+    id("publish-conventions")
+}
+
+// MPP projects pack their sources automatically, java libraries need to explicitly pack them
+val sourcesJar = tasks.register<Jar>("sourcesJar") {
```
adam-enko, line 8:
```suggestion
val sourcesJar by tasks.registering(Jar::class) {
```

using `by tasks.registering` means the name of the task is set once, in the property name, rather than using a string.

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
```diff
@@ -0,0 +1,42 @@
+plugins {
+    `maven-publish`
+    signing
+}
+
+// empty xxx-javadoc.jar
+val javadocJar = tasks.register<Jar>("javadocJar") {
```
adam-enko, line 7:
```suggestion
val javadocJar by tasks.registering(Jar::class) {
```

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+    archiveClassifier.set("javadoc")
+}
+
+publishing {
+    repositories {
+        configureMavenPublication(this, project)
```
adam-enko, line 13: The `configureMavenPublication()` util function is only used here - I think it'd be easier to copy-paste the function body here, and delete the util function.

WDYT?

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    publications.withType<MavenPublication>().all {
+        pom.configureMavenCentralMetadata(project)
```
adam-enko, line 17: Same comment as with `configureMavenPublication()` above -  this is th only usage of the `configureMavenCentralMetadata()` util. Could it be inlined?

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+
+        // Task ':linkDebugTest<platform>' uses this output of task ':sign<platform>Publication' without declaring an explicit or implicit dependency
+        tasks.findByName("linkDebugTest$pubName")?.configure<Task> {
+            mustRunAfter(this@sign)
+        }
+        // Task ':compileTestKotlin<platform>' uses this output of task ':sign<platform>Publication' without declaring an explicit or implicit dependency
+        tasks.findByName("compileTestKotlin$pubName")?.configure<Task> {
+            mustRunAfter(this@sign)
+        }
+    }
```
adam-enko, line 41:
```suggestion
    // NOTE: This is a temporary WA, see KT-61313.
    tasks.withType<Sign>().configureEach sign@{
        val pubName = name.substringAfter("sign").substringBefore("Publication")

        // Task ':linkDebugTest<platform>' uses this output of task ':sign<platform>Publication' without declaring an explicit or implicit dependency
        tasks.findByName("linkDebugTest$pubName")?.configure<Task> {
            mustRunAfter(this@sign)
        }
        // Task ':compileTestKotlin<platform>' uses this output of task ':sign<platform>Publication' without declaring an explicit or implicit dependency
        tasks.findByName("compileTestKotlin$pubName")?.configure<Task> {
            mustRunAfter(this@sign)
        }
    }
```

I think that the first workaround isn't necessary, the second one should work.

buildSrc/src/main/kotlin/non-multiplatform.publish-conventions.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+            create<MavenPublication>("pluginMaven") {
+                artifact(sourcesJar)
+            }
+        } else {
+            create<MavenPublication>("maven") {
+                from(components["java"])
+                artifact(sourcesJar)
+                if (project.name.endsWith("-maven-plugin")) {
+                    pom.packaging = "maven-plugin"
+                }
```
adam-enko, line 26: It'd make sense to move this subproject-specific logic into the subproject's build.gradle. Convention plugins should (ideally) be unopinionated, and avoid having subproject specific logic.

adam-enko (changes requested):

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
dkrasnoff: Actually I'm not sure that we don't need this:
```
 tasks.withType<AbstractPublishToMaven>().configureEach {
        dependsOn(tasks.withType<Sign>())
    }
 ```
 because the second workaround adds `dependsOn` only for mpp test tasks, but we add this plugin for all maven publications.

Comment (adam-enko) [reply buildSrc/src/main/kotlin/publish-conventions.gradle.kts]: They're both workarounds for this issue https://github.com/gradle/gradle/issues/26132

The error `Some problems were found with the configuration of task [...] Task ':linkDebugTestLinuxX64' uses this output of task ':signLinuxX64Publication' without declaring an explicit or implicit dependency` only really occurs in Kotlin Multiplatform projects. 

This workaround will work, but it isn't optimal:

```kotlin
tasks.withType<AbstractPublishToMaven>().configureEach {
    dependsOn(tasks.withType<Sign>())
}
```

It causes _all_ signing tasks to run, even when if they might not be necessary.

This workaround is better:

```kotlin
tasks.withType<AbstractPublishToMaven>().configureEach {
    mustRunAfter(tasks.withType<Sign>())
}
```

Now the signing tasks will only run if required (as normal), but there's no chance of overlapping outputs because Gradle will make sure _all_ signing tasks run before _any_ publishing tasks.

The other workaround is even more optimised, and will only add the `mustRunAfter()` fix to the specifically affect KMP tasks.

I think that the `mustRunAfter(tasks.withType<Sign>())` workaround is a good compromise, and we should choose it. It's simple, and just works, without having to do string-based checks (which can be flakey). It's not as optimised, but it's not a big deal.

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
dkrasnoff: I meet such error:
```
> Cannot convert the provided notation to an object of type MavenArtifact: org.gradle.kotlin.dsl.RegisteringDomainObjectDelegateProviderWithTypeAndAction@9a7d1ac.
  The following types/formats are supported:
    - Instances of MavenArtifact.
    - Instances of AbstractArchiveTask, for example jar.
    - Instances of PublishArtifact
    - Instances of Provider
    - Maps containing a 'source' entry, for example [source: '/path/to/file', extension: 'zip'].
    - Anything that can be converted to a file, as per Project.file()
```
in 
```
        if (name != "kotlinMultiplatform") { // The root module gets the JVM's javadoc JAR
            artifact(javadocJar)
        }
```
with such approach

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
dkrasnoff: I get it, there was no:
```
tasks.withType<AbstractPublishToMaven>().configureEach {
    mustRunAfter(tasks.withType<Sign>())
}
```
in suggested changes, so I thought you just removed it

buildSrc/src/main/kotlin/non-multiplatform.publish-conventions.gradle.kts
dkrasnoff: rewrote

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
dkrasnoff: rewrote

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
dkrasnoff: copy-pasted

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
dkrasnoff: inlined

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
dkrasnoff: I've applied:
```
tasks.withType<AbstractPublishToMaven>().configureEach {
    mustRunAfter(tasks.withType<Sign>())
}
```

buildSrc/src/main/kotlin/non-multiplatform.publish-conventions.gradle.kts
dkrasnoff: Moved + rewrote `build.gradle` to `build.gradle.kts` in `kotlinx-atomicfu.atomicfu-maven-plugin`, thus could you please have a look on this change too

buildSrc/src/main/kotlin/non-multiplatform.publish-conventions.gradle.kts
```diff
@@ -5,27 +5,16 @@ plugins {
...
-val sourcesJar = tasks.register<Jar>("sourcesJar") {
+val sourcesJar by tasks.registering(Jar::class) {
     archiveClassifier.set("sources")
     from("src/main/kotlin")
 }
-
 publishing {
     // Configure java publications for non-MPP projects
     publications {
-        if (project.name == "atomicfu-gradle-plugin") {
```
dkrasnoff: I've just removed this logic, because `from(components["java"])` does not fail, even without `java` component, as I see

dkrasnoff:
Depends on https://github.com/Kotlin/kotlinx-atomicfu/pull/434

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
```diff
@@ -0,0 +1,75 @@
...
+                url = "https://github.com/Kotlin/kotlinx.atomicfu"
+            }
+        }
+
+
+        signPublicationIfKeyPresent(project, this)
+        // add empty javadocs
+        if (name != "kotlinMultiplatform") { // The root module gets the JVM's javadoc JAR
+            artifact(javadocJar)
+        }
```
adam-enko, line 55: It'd be nice to tidy up this config in a later PR.

```suggestion
        // TODO https://github.com/Kotlin/kotlinx-atomicfu/issues/421
        //      Figure out if there's an alternative to string-based config
        //      (if the name changes this config will silently stop working).
        //      Maybe the Javadoc JAR can always be added?
        // add empty javadocs
        
        if (name != "kotlinMultiplatform") { // The root module gets the JVM's javadoc JAR
            artifact(javadocJar)
        }
```

adam-enko (changes requested):

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
```diff
@@ -0,0 +1,75 @@
...
+
+        // Task ':linkDebugTest<platform>' uses this output of task ':sign<platform>Publication' without declaring an explicit or implicit dependency
+        tasks.findByName("linkDebugTest$pubName")?.let { linkTask ->
+            linkTask.mustRunAfter(this@sign)
+        }
+        // Task ':compileTestKotlin<platform>' uses this output of task ':sign<platform>Publication' without declaring an explicit or implicit dependency
+        tasks.findByName("compileTestKotlin$pubName")?.let { compileTask ->
+            compileTask.mustRunAfter(this@sign)
+        }
+    }
```
adam-enko, line 74: Continued from https://github.com/Kotlin/kotlinx-atomicfu/pull/435#discussion_r1598070977

```suggestion
    // NOTE: This is a temporary WA, see KT-61313.
    // Task ':linkDebugTest<platform>' uses this output of task ':sign<platform>Publication' without declaring an explicit or implicit dependency
    tasks.withType<AbstractPublishToMaven>().configureEach {
        mustRunAfter(tasks.withType<Sign>())
    }
```

Let's use a simpler workaround. Even if it's less efficient it's easier to understand & maintain.

buildSrc/src/main/kotlin/publish-conventions.gradle.kts
adam-enko: https://github.com/Kotlin/kotlinx-atomicfu/pull/435/files#r1610189771

buildSrc/src/main/kotlin/non-multiplatform.publish-conventions.gradle.kts
```diff
@@ -0,0 +1,20 @@
+plugins {
+//     Regular java modules need 'java-library' plugin for proper publication
+    `java-library`
+    id("publish-conventions")
+}
+
+// MPP projects pack their sources automatically, java libraries need to explicitly pack them
+val sourcesJar by tasks.registering(Jar::class) {
+    archiveClassifier.set("sources")
+    from("src/main/kotlin")
```
adam-enko, line 10: If this convention assumes there's a `src/main/kotlin` directory shouldn't it apply the Kotlin/JVM plugin? 🤔

atomicfu-maven-plugin/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+            "--settings", "./settings.xml",
+            "--errors",
+            "--batch-mode",
+            "--file", generatePomFileForMavenPublication.destination.toString(),
+            "org.apache.maven.plugins:maven-plugin-plugin:3.5.1:descriptor"
+        )
+    )
+    commandLine = args
+    doLast {
+        val descriptorFile = pluginDescriptorFile.get().getAsFile()
```
adam-enko, line 53:
```suggestion
        val descriptorFile = pluginDescriptorFile.get().asFile
```

> Use of getter method instead of property access syntax

atomicfu-maven-plugin/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+        assert(descriptorFile.exists(), { "$descriptorFile: was not generated" })
```
adam-enko, line 54:
```suggestion
        assert(descriptorFile.exists()) { "$descriptorFile: was not generated" }
```

> Lambda argument should be moved out of parentheses

adam-enko: Also, `assert()` won't work if Java assertions are disabled. Could use `require()` instead?

```suggestion
        require(descriptorFile.exists(), { "$descriptorFile: was not generated" })
```

adam-enko:
> @adam-enko, CI say we can not get rid of:
> 
> ```
>    // NOTE: This is a temporary WA, see KT-61313.
>     tasks.withType<Sign>().configureEach sign@{
> ```

D'oh! Oh well, better revert it. Thanks for trying!

mvicsokolova (approved):

PR #436, opened 2024-05-10
Removed allprojects block from root build script

dkrasnoff:
Gradle docs says that we should avoid [allprojects and subprojects blocks](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration).
In this PR I made next steps:
- Moved common logic from `allprojects` block in the root build script to `common-conventions.gradle.kts`
- Moved multiplatform connected logic from `allprojects` block in the root build script to `multiplatform-conventions.gradle.kts`
- Rewrote to kotlin dsl build script in `atomicfu-gradle-plugin`

Depends on

* https://github.com/Kotlin/kotlinx-atomicfu/pull/435
* https://github.com/Kotlin/kotlinx-atomicfu/pull/434


buildSrc/src/main/kotlin/common-conventions.gradle.kts
```diff
@@ -0,0 +1,18 @@
...
+    compilerOptions { freeCompilerArgs.add("-Xexpect-actual-classes") }
+}
+
+plugins.withType<JavaPlugin> {
+    configure<JavaPluginExtension> {
+        toolchain {
+            languageVersion.set(JavaLanguageVersion.of(8))
+        }
+    }
+}
```
adam-enko, line 18: These opt-in and toolchain conventions are specific to Kotlin subprojects.

I think it'd be better to do the following:

1. Move the KMP-specific expect/actual suppression to the KMP convention
2. Create a `kotlin-jvm` convention with the JVM toolchain config.

buildSrc/src/main/kotlin/common-conventions.gradle.kts
adam-enko, line 1: Instead of applying `common-conventions` to each subproject, I think it'd be better to apply `common-conventions` to the other convention plugins.

For example, add `plugins { id("common-conventions") }` to `multiplatform-conventions.gradle.kts`.


The `plugins {}` block in each subproject would then be simpler, since it would have a 'main' convention.

buildSrc/src/main/kotlin/multiplatform-conventions.gradle.kts
adam-enko, line 1: I think the 'multiplatform-conventions' name is a bit generic. When I'm looking at a `build.gradle` it'd help me to be able to see a 'kotlin' plugin, and then I know that the subproject is Kotlin-based. 'multiplatform' doesn't have the same impact.

What do you think about renaming it to `kotlin-multiplatform-convention`?

adam-enko (changes requested):

dkrasnoff:
Depends on https://github.com/Kotlin/kotlinx-atomicfu/pull/435

buildSrc/src/main/kotlin/common-conventions.gradle.kts
dkrasnoff: fixed

buildSrc/src/main/kotlin/multiplatform-conventions.gradle.kts
dkrasnoff: fixed

build.gradle
```diff
@@ -15,39 +11,6 @@ plugins {
...
-    if (deployVersion != null) project.version = deployVersion
-
-    // this fixes "org.gradle.jvm.version" in Gradle metadata
-    plugins.withType(JavaPlugin).configureEach {
-        java {
-            toolchain {
-                languageVersion.set(JavaLanguageVersion.of(8))
-            }
-        }
-    }
```
adam-enko, line 35: Just to double check, is the language version config still present in the new plugins?

dkrasnoff: I didn't remove it, just changed to:
```
kotlin {
    jvmToolchain(8)
}
```
I've checked it by comparing published artifacts and artifacts before changes and after are equal. Or should I check it other way?

adam-enko: Great, thanks for checking. I just wanted to confirm, no further action needed 👍

adam-enko (approved):
(I meant to approve before, but I guess I missed the approve button) 

A small note: `compile-options-conventions.gradle.kts` should be converted to a `kotlin-base-conventions..gradle.kts` and applied in the `kotlin-jvm` and `kotlin-multiplatform` conventions. But that can be done in a later PR - I'll leave it up to you.

PR #445, opened 2024-06-27
Fix the new ID of the atomicfu-gradle-plugin

mvicsokolova:
To be published on GPP the gradle plugin should be namespaced and the plugin cannot be published with the old id `kotlinx-atomicfu` as it was before. So, the plugin name is changed to `org.jetbrains.kotlinx.atomicfu`. 

And here is how the plugin should be applied now: 

```
plugins {
     id("org.jetbrains.kotlinx.atomicfu") version "0.25.0"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,28 +112,53 @@ operations. They can be also atomically modified via `+=` and `-=` operators.
 ### Apply plugin
 #### Gradle configuration
 
-In top-level build file:
+> **New plugin id:** Please pay attention, that starting from version `0.25.0` the plugin id is `org.jetbrains.kotlinx.atomicfu`
+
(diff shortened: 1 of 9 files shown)


gradle.properties
```diff
@@ -2,7 +2,7 @@
 # Copyright 2017-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
 #
 
-version=0.24.0-SNAPSHOT
+version=0.24.0-SNAPSHOT-publishOnce
```
dkrasnoff, line 5: publishOnce

dkrasnoff (approved):

PR #446, opened 2024-06-28
Version 0.25.0

mvicsokolova:


README.md
```diff
@@ -263,7 +288,7 @@ IR for all the target backends:
 
 To turn on IR transformations set the following properties in your `gradle.properties` file:
 
-> Please note, that starting from version `0.24.0` of the library your project is required to use `Kotlin version >= 1.9.0`.
```
fzhinkin, line 266: Here, the version should remain 0.24.0, innit?

README.md
```diff
@@ -273,7 +298,7 @@ kotlinx.atomicfu.enableJsIrTransformation=true // for JS IR transformation
 ```
 
 <details open>
-<summary> Here are the configuration properties in case you use older versions of the library lower than 0.24.0. </summary>
```
fzhinkin, line 276: And here

README.md
mvicsokolova: Thank you! I didn't notice that

PR #451, opened 2024-07-02
Build config/build scans and build cache

dkrasnoff:
In this commit added:

- [Build Scans](https://docs.gradle.org/current/userguide/build_scans.html), which allows you to find more information about the build in [Jetbratins develocity](https://ge.jetbrains.com/scans?search.rootProjectNames=kotlinx-atomicfu)
- [Build cache](https://docs.gradle.org/current/userguide/build_cache.html), which allows builds to reuse outputs produced by other builds


README.md
```diff
@@ -475,3 +475,19 @@ See [gradle.properties](gradle.properties) in AtomicFU project for its `kotlin_v
 
 Available Kotlin/Native targets are based on non-deprecated official targets [Tier list](https://kotlinlang.org/docs/native-target-support.html)
  with the corresponding compatibility guarantees.
+
+### Gradle Build Scans
+
+[Gradle Build Scans](https://scans.gradle.com/) can provide insights into a Atomicfu Build.
```
adam-enko, line 481:
```suggestion
[Gradle Build Scans](https://scans.gradle.com/) can provide insights into an Atomicfu Build.
```

nitpick typo

build-settings-logic/src/main/kotlin/Utils.kt
adam-enko, line 1: These utils are only used in `gradle-build-scan.settings.gradle.kts` - maybe they should be moved into the same file?

Alternatively, I think a more descriptive name would be nice. What about `BuildScanProperties.kt`?

build-settings-logic/src/main/kotlin/Utils.kt
```diff
@@ -0,0 +1,31 @@
...
+        atomicfuProperty("build.scan.enabled", String::toBoolean)
+            .orElse(buildingOnCi)
+
+/**
+ * Optionaly override the default name attached to a Build Scan.
+ */
+val Settings.buildScanUsername: Provider<String>
+    get() =
+        atomicfuProperty("build.scan.username")
+            .map(String::trim)
```
adam-enko, line 24: `buildScanUsername` should have a `<default>` value, or else the `overriddenName == "<default>" -> originalUsername` branch won't be triggered in `obfuscation {}`?

```suggestion
val Settings.buildScanUsername: Provider<String>
    get() =
        atomicfuProperty("build.scan.username")
            .orElse("<default>")
            .map(String::trim)
```

gradle.properties
```diff
@@ -16,4 +16,6 @@ systemProp.org.gradle.internal.publish.checksums.insecure=true
 org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=2g
 
 # we can remove this flag, when move to kotlin-gradle-plugin `2.0.0+`
-kotlin.native.distribution.downloadFromMaven=true
\ No newline at end of file
+kotlin.native.distribution.downloadFromMaven=true
+org.gradle.caching=true
```
adam-enko, line 20: 🎉

settings.gradle.kts
```diff
@@ -105,6 +104,13 @@ dependencyResolutionManagement {
     }
 }
 
+rootProject.name = "kotlinx-atomicfu"
+
+plugins {
+    id("gradle-build-scan")
+    id("gradle-build-cache")
+}
```
adam-enko, line 112: It's not immediately obvious that these plugins are Atomicfu convention plugins. It'd be nice to give them a descriptive namespace. E.g. add `package atomicfu_build` to the top of the `.kts` files (but keep the files the same location)

build-settings-logic/settings.gradle.kts
```diff
@@ -0,0 +1,18 @@
...
+    }
+}
+dependencyResolutionManagement {
+
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+        }
+    }
+}
```
adam-enko, line 16:
```suggestion
dependencyResolutionManagement {
    repositories {
        mavenCentral()
        gradlePluginPortal()
    }

    versionCatalogs {
        create("libs") {
            from(files("../gradle/libs.versions.toml"))
        }
    }
}
```

build-settings-logic/src/main/kotlin/gradle-build-scan.settings.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+                        !overriddenName.isNullOrBlank() -> overriddenName
+                        overriddenName == "<default>" -> originalUsername
+                        else -> "unknown"
+                    }
+                }
+            }
+        }
+    }
+
+}
```
adam-enko, line 35:
```suggestion
    }
}
```

formatting nitpick

adam-enko (approved):
Great work! Exciting to see the build config shaping up really well.

build-settings-logic/src/main/kotlin/Utils.kt
dkrasnoff: I would leave it as is for the future, if we have some shared logic, it could be placed in `Utils.kt` file

mvicsokolova (approved):

mvicsokolova:
Looks great, thank you!)

PR #453, opened 2024-07-11
Enable cache-redirector

dkrasnoff:
Cache-redirector works like CDN for JB builds.
The major part of the code was copy-stolen from [kotlin repo's](https://github.com/JetBrains/kotlin/blob/6c3471c8b5a1ca2871b990363f481be1bfcb2225/repo/scripts/cache-redirector.settings.gradle.kts#L4) cache-redirector (it worth to add to a shared build-scripts for JB, when such repository will be exist). 

In this PR I also removed duplicated properties for setting koltin version and custom repository URL according to our checklist #421


integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Utils.kt
```diff
@@ -24,11 +28,52 @@ internal fun GradleBuild.buildGradleByShell(
     return BuildResult(exitCode, logFile)
 }
 
+internal fun Path.enableCacheRedirector() {
+    // Path relative to the current gradle module project dir
+    val redirectorScript =
+        Paths.get("../build-settings-logic/src/main/kotlin/atomicfu-cache-redirector.settings.gradle.kts")
+            .toAbsolutePath()
+            .normalize()
+            .toFile()
```
adam-enko, line 37: For Gradle up-to-date checks this file should be registered as a test task input.

dkrasnoff: I'm unsure if it is necessary because the cache-redirector is only responsible for downloading dependencies.
When the task is up-to-date, we don't need to download dependencies, but if we change smth in cache-redirector, we also don't need to rerun the task. Or am I wrong?

dkrasnoff: also, [here](https://github.com/Kotlin/kotlinx-atomicfu/blob/dc1c5f5162dd976a9bded61d0d4b967c05d719ae/integration-testing/build.gradle.kts#L91) we have 
```kotlin
outputs.upToDateWhen { false }
```
which as I understand always forces to run `functionalTest`

mvicsokolova (approved):

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Utils.kt
dkrasnoff: moved path to system property

build-settings-logic/src/main/kotlin/atomicfu-cache-redirector.settings.gradle.kts
```diff
@@ -0,0 +1,202 @@
...
+fun RepositoryHandler.redirect() = configureEach {
+    when (this) {
+        is MavenArtifactRepository -> url = url.maybeRedirect()
+        is IvyArtifactRepository -> @Suppress("SENSELESS_COMPARISON") if (url != null) {
+            url = url.maybeRedirect()
+        }
+    }
+}
+
+pluginManagement.repositories.redirect()
```
dkrasnoff, line 195: @adam-enko honestly say, I still don't understand, why do you suggest to move it to gradle dir. It looks like a real settings plugin, which should be applied directly to `settings.gradle.kts`.
I would suggest not to move it.

adam-enko: Well, because it's not a real settings plugin :) It's applied as a file-based script plugin in `build-settings-logic/settings.gradle.kts` and in all of the integration tests.

While technically it can be both a file-based and pre-compiled script plugin, it's a little odd. I also don't like that it's easy to accidentally break the the plugin by using a utility function from Utils.kt.

That said, if it works, it works. I'm not going to block the PR because of it.

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Utils.kt
```diff
@@ -24,11 +28,56 @@ internal fun GradleBuild.buildGradleByShell(
...
+        System.getProperty("cache.redirector.path").ifBlank { null } ?: error("Cache redirector has not been set")
+
+    // Path relative to the current gradle module project dir
+    val redirectorScript =
+        Paths.get(cacheRedirectorPath)
+            .toAbsolutePath()
+            .normalize()
+            .toFile()
+
+    val gradleDir = resolve("gradle").also { it.createDirectories() }
```
adam-enko, line 43:
```suggestion
    val gradleDir = resolve("gradle").createDirectories()
```

nitpick: `createDirectories()` returns the same value, so `also {}` could be removed.

adam-enko (approved):

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/runner/Utils.kt
dkrasnoff: moved this change to https://github.com/Kotlin/kotlinx-atomicfu/pull/460

PR #456, opened 2024-07-18
Build config/configuration cache

dkrasnoff:


gradle.properties
```diff
@@ -17,3 +17,4 @@ org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=2g
 
 org.gradle.caching=true
 
+org.gradle.configuration-cache=true
```
adam-enko, line 20: 🚀

buildSrc/src/main/kotlin/MavenPomUtils.kt
adam-enko, line 1: These utilities could be `internal`? I assume they're only used in buildSrc?

atomicfu-maven-plugin/build.gradle.kts
```diff
@@ -13,31 +19,84 @@ dependencies {
     api(libs.bundles.maven)
 }
 
-val outputDir = tasks.compileKotlin.get().destinationDirectory
+val outputDir = tasks.compileKotlin.get().destinationDirectory.get().asFile
```
adam-enko, line 22: This looks suspicious to me. Generally we should try to avoid `.get()`, because it will eagerly configure a task. Also, `get().asFile` will make Gradle forget the task link.

buildSrc/src/main/kotlin/MavenPomUtils.kt
dkrasnoff: no in atomicfu-maven-plugin/build.gradle.kts as well

buildSrc/src/main/kotlin/MavenPomUtils.kt
```diff
@@ -0,0 +1,47 @@
...
+
+fun MavenPom.configureMavenPluginPomAttributes(projectName: String) {
+    name.set(projectName)
+    packaging = "maven-plugin"
+    description.set("Atomicfu Maven Plugin")
+}
+
+fun Document.getElement(tagName: String): Node =
+    getElementsByTagName(tagName).item(0)
+        ?: error("No element named '$tagName' in Document $this")
```
adam-enko, line 19: Is this unused? Can it be deleted?

buildSrc/src/main/kotlin/MavenPomUtils.kt
```diff
@@ -0,0 +1,47 @@
...
+
+fun Element.getElement(tagName: String): Node =
```
adam-enko, line 21:
```suggestion
internal fun Element.getElement(tagName: String): Node =
```

buildSrc/src/main/kotlin/MavenPomUtils.kt
adam-enko: Okay, I see, thanks. It'd be nice to move the POM configuration into a single util function, but that's not relevant to this PR.

buildSrc/src/main/kotlin/MavenPomUtils.kt
```diff
@@ -0,0 +1,47 @@
+@file:JvmName("MavenPomConfiguration")
+
+import org.gradle.api.publish.maven.*
```
adam-enko, line 3:
```suggestion
import org.gradle.api.publish.maven.*
```

Since all the build scripts are Kotlin now, the JvmName could be removed?

buildSrc/src/main/kotlin/common-conventions.gradle.kts
```diff
@@ -1,5 +1,2 @@
 val deployVersion = project.findProperty("DeployVersion")
```
adam-enko, line 1:
```suggestion
val deployVersion: String? = project.findProperty("DeployVersion")?.toString()?.ifBlank { null }
```

It's unrelated to this PR, but I'd recommend being defensive here, just in case a plain-string gets passed.

I'll leave it up to you whether to implement or not, but I wanted to mention it while it's on my mind.

buildSrc/src/main/kotlin/common-conventions.gradle.kts
adam-enko, line 1: I think it's a bit unusual that this file only has two lines... It could be merged into `base-publish-conventions.gradle.kts`? The version only really matters for published artifacts.

Again, it's not related to CC, so I'll just mention it and leave it up to you.

atomicfu-maven-plugin/build.gradle.kts
dkrasnoff: Fixed

buildSrc/src/main/kotlin/MavenPomUtils.kt
dkrasnoff: Removed

buildSrc/src/main/kotlin/MavenPomUtils.kt
dkrasnoff: Applied

buildSrc/src/main/kotlin/MavenPomUtils.kt
dkrasnoff: Applied

buildSrc/src/main/kotlin/common-conventions.gradle.kts
dkrasnoff: Thanks, applied

buildSrc/src/main/kotlin/common-conventions.gradle.kts
dkrasnoff: Thanks, merged into one convention plugin

adam-enko (approved):

mvicsokolova (approved):

PR #460, opened 2024-07-23
De-duplicated the version override in the actual project and in buildSrc

dkrasnoff:
This PR adds common settings plugin to: 
- root `settings.gradle.kts` 
- buildSrc `settings.gradle.kts` 

It removes duplicated logic by setting custom Kotlin version and custom repository URL. (`kotlin_version`, `kotlin_repo_url`, `build_snapshot_train`).

Should be checked after https://github.com/Kotlin/kotlinx-atomicfu/pull/453.

### DIFF
--- a/build-settings-logic/src/main/kotlin/atomicfu-dependency-resolution-management.settings.gradle.kts
+++ b/build-settings-logic/src/main/kotlin/atomicfu-dependency-resolution-management.settings.gradle.kts
@@ -1,0 +1,79 @@
+import org.gradle.kotlin.dsl.maven
+
+pluginManagement {
+    repositories {
+        mavenCentral()
+        gradlePluginPortal()
+
+        val additionalRepositoryProperty = providers.gradleProperty("kotlin_repo_url")
+        if (additionalRepositoryProperty.isPresent) {
+            maven(additionalRepositoryProperty.get()) {
+                name = "KotlinDevRepo"
+            }
+            logger.info("A custom Kotlin repository ${additionalRepositoryProperty.get()} was added")
+        }
+
+        maven("https://oss.sonatype.org/content/repositories/snapshots") {
+            mavenContent { snapshotsOnly() }
+        }
+    }
+}
+
+dependencyResolutionManagement {
+
+    @Suppress("UnstableApiUsage")
+    repositoriesMode.set(RepositoriesMode.PREFER_SETTINGS)
+
+    @Suppress("UnstableApiUsage")
+    repositories {
+        maven("https://oss.sonatype.org/content/repositories/snapshots") {
+            mavenContent { snapshotsOnly() }
+        }
+
+        val additionalRepositoryProperty = providers.gradleProperty("kotlin_repo_url")
+        if (additionalRepositoryProperty.isPresent) {
+            maven(additionalRepositoryProperty.get()) {
+                name = "KotlinDevRepo"
+            }
+            logger.info("A custom Kotlin repository ${additionalRepositoryProperty.get()} was added")
+        }
+
+        mavenCentral()
+
+
+        // we have such a task https://youtrack.jetbrains.com/issue/KT-34732 to move these artifacts to the Maven repository,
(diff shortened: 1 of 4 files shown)


build-settings-logic/src/main/kotlin/atomicfu-dependency-resolution-management.settings.gradle.kts
```diff
@@ -0,0 +1,87 @@
...
+        /*
+         * This property group is used to build kotlinx.atomicfu against Kotlin compiler snapshots.
+         * When build_snapshot_train is set to true, mavenLocal and Sonatype snapshots are added to a repository list
+         * (the former is required for AFU and public, the latter is required for compiler snapshots).
+         * DO NOT change the name of these properties without adapting the kotlinx.train build chain.
+         */
+        val buildSnapshotTrainGradleProperty = providers.gradleProperty("build_snapshot_train")
+        if (buildSnapshotTrainGradleProperty.isPresent) {
+            maven(url = uri("https://oss.sonatype.org/content/repositories/snapshots"))
+        }
```
adam-enko, line 23: It's usually better to keep the build config as stable as possible. This avoids surprises, and helps with Gradle caching (configuration cache in this case).

[There was an issue a couple of weeks ago caused by conditional Maven repos](https://jetbrains.slack.com/archives/C4U955N6B/p1721204053175179).

Could the Maven Central snapshots be added without the conditional logic? The content could be filtered to help performance, so that Gradle will only look for snapshots in it.

```suggestion
        maven("https://oss.sonatype.org/content/repositories/snapshots") {
            mavenContent { snapshotsOnly }
        }
```

build-settings-logic/src/main/kotlin/atomicfu-dependency-resolution-management.settings.gradle.kts
```diff
@@ -0,0 +1,87 @@
...
+        * DO NOT change the name of these properties without adapting kotlinx.train build chain.
+        */
+        val buildSnapshotTrainGradleProperty = providers.gradleProperty("build_snapshot_train")
+        if (buildSnapshotTrainGradleProperty.isPresent) {
+            maven(url = uri("https://oss.sonatype.org/content/repositories/snapshots"))
+        }
+
+        val additionalRepositoryProperty = providers.gradleProperty("kotlin_repo_url")
+        if (additionalRepositoryProperty.isPresent) {
+            maven(url = uri(additionalRepositoryProperty.get()))
```
adam-enko, line 47: The kotlin-dsl has a shorter method of adding repositories, which just accepts a string.

I also think it's a nice idea to add a name to the repositories, so that error messages and Build Scan reports are more human readable (otherwise Gradle generates a non-specific name).

```suggestion
            maven(additionalRepositoryProperty.get()) {
                name = "KotlinDevRepo"
            }
```

adam-enko (approved):

build-settings-logic/src/main/kotlin/atomicfu-dependency-resolution-management.settings.gradle.kts
dkrasnoff: I see, didn't think about Build Scan reports. It makes sense, thanks!

build-settings-logic/src/main/kotlin/atomicfu-dependency-resolution-management.settings.gradle.kts
dkrasnoff: @adam-enko I started exploring this idea and came up to an interesting conclusion, that `https://oss.sonatype.org/content/repositories/snapshots` is not as it was expected. Instead, in the train builds, we use only Maven local to publish intermediate artifacts.

adam-enko: Ah okay, so Maven Central Snapshots is not used at all? If so, then it makes sense to remove it, *unless* it might be used in the future.

mvicsokolova (approved):

PR #467, opened 2024-08-14
Set server on develocity extension

CristianGM:
This way we can use the Develocity token to authenticate on CI automatically


build-settings-logic/src/main/kotlin/atomicfu-gradle-build-scan.settings.gradle.kts
```diff
@@ -7,8 +7,8 @@ plugins {
 develocity {
     if (buildScanEnabled.get()) {
         val overriddenName = buildScanUsername.orNull
+        server = "https://ge.jetbrains.com/"
```
dkrasnoff, line 10: nit: maybe it worth to left a comment, why this line here and not in `buildScan` for future generations

dkrasnoff:
Thanks a lot! 👍

Issue #462, opened 2024-07-31
AtomicFu QoS Problems on iOS


stefanhaustein:
## Summary

On Darwin (iOS) the kernel must support a lock to handle QoS properly, and a custom built lock using atomics – such as the one in [synchronized.kt](https://source.corp.google.com/piper///depot/google3/third_party/kotlin/kotlinx_atomicfu/atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt)  – won't fit that model (atomics do not handle QoS)

## Background

[iOS resolves priority inversions between threads by temporarily promoting the QoS of the lower priority thread to the same level as the higher priority thread being blocked](https://developer.apple.com/library/archive/documentation/Performance/Conceptual/EnergyGuide-iOS/PrioritizeWorkWithQoS.html). In order for this promotion to occur, you must use one of the Apple-supplied lock mechanisms that support QoS donation.

Not all locks available in Apple APIs support QoS. For example, spinlocks, semaphores, and condition locks (including NSCondition) do not. Apple has aggressively deprecated their spinlock APIs, and has privately warned us not to use condition locks in performance-sensitive code.

Unfortunately, while the QoS-compatibility of locks has evolved to "common knowledge" in iOS dev, there is no authoritative Apple documentation.

Some public discussions of locks and QoS:

- https://threadreaderapp.com/thread/1229999590482444288.html
- https://developer.apple.com/library/archive/documentation/Performance/Conceptual/EnergyGuide-iOS/PrioritizeWorkWithQoS.html 
- https://hacks.mozilla.org/2022/10/improving-firefox-responsiveness-on-macos/ 
- https://openradar.appspot.com/23896366 
- https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20151214/000345.html 
- https://www.mikeash.com/pyblog/friday-qa-2017-10-27-locks-thread-safety-and-swift-2017-edition.html 
- https://mjtsai.com/blog/2015/12/16/osspinlock-is-unsafe/ 

The takeaway of these discussions is that it is possible for a thread or serial queue (these are not exactly the same thing) to block for too long, or even indefinitely, on a lock that does not perform QoS promotion. The most problematic case is one where the low-priority thread holds the lock, but is never serviced because there are always higher QoS threads ready to run but blocking on the lock. Spinlocking is apparently particularly bad at triggering this case.

[...]

PR #468, opened 2024-08-26
Use pthread_mutex instead of a custom lock on Apple platforms

stefanhaustein:
Use pthread_mutex instead of a custom lock on Apple platforms, as for Darwin the kernel must support a lock to handle QoS properly, and a custom built lock using atomics – such as the one currently used in nativeMain/synchronized.kt – won't fit that model (atomics do not handle QoS)

See issue #462 for details.

This PR introduces a new apple target and splits nativeNonApple (with the custom lock) out of native, keeping the dependency on the remaining native code in both cases.


stefanhaustein:
P.S.: I accidentally included .gitignore for local settings (should be fine) and some jre fetcher in settings... Should I re-generate?

mvicsokolova:
Thank you! Looks good to me as the first solution 👍🏻
I just need to rebase these changes onto `develop` branch, as it contains a bunch of changes in the build scripts

stefanhaustein:
Thanks! It probably makes sense to push the mutex pool up to the shared native part then?

qwwdfsad:
>Does the pool make any sense here? T

Do you mean in lock implementation in general?
It does because otherwise lock allocation performance model drastically differs between platforms and we would like to avoid that. E.g. we know that lock allocation is cheap and do so in coroutines left and right

stefanhaustein:
I have an updated implementation incorporating feedback here and from compose-kmp... Given the amount of changes, the simplest way forward seems to be to close this and to create a new PR after I have synced my fork.

PR #470, opened 2024-09-05
Fix wrong test setup

Tapchicoma:
Test project was applying Kotlin KMP plugin in subprojects, but not in the root 'build.gradle.kts' which is wrong from KMP point of view as it introduces Gradle classpath isolation problem.

### DIFF
--- a/integration-testing/examples/multi-module-test/build.gradle.kts
+++ b/integration-testing/examples/multi-module-test/build.gradle.kts
@@ -1,0 +1,3 @@
+plugins {
+    kotlin("multiplatform") version libs.versions.kotlinVersion.get() apply false
+}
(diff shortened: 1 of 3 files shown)


Tapchicoma:
Fixes failure in MR into `2.0.21` branch for https://youtrack.jetbrains.com/issue/KT-70700

mvicsokolova (approved):

woainikk:
can we merge it directly to develop branch?

mvicsokolova:
Yes, we should first merge it into develop and then rebase kotlin-community

mvicsokolova:
@Tapchicoma could you please rebase this commit onto develop? Or should I better do this?

Tapchicoma:
Actually I've just rebased against `develop`

woainikk (approved):

Issue #431, opened 2024-04-28
Remove 


mvicsokolova:

lppedd:
Other than performance, we should also consider bundle size.  
Kotlin's stdlib already adds quite a bit to the total size, so if libraries can shave off some KBs, it's always nice.

PR #475, opened 2024-10-07
Add api kotlinx-atomicfu dependency for JS and WASM targets.

### LINKED ISSUE #448: Warning about compileOnly dependency with Kotlin 2.0.20-Beta1
When building [a project](https://github.com/kordlib/kord) that uses AtomicFU with Kotlin 2.0.20-Beta1, I get this warning:

```
w: A compileOnly dependency is used in targets: Kotlin/JS.
Dependencies:
    - org.jetbrains.kotlinx:atomicfu:0.25.0 (source sets: jsMain)

Using compileOnly dependencies in these targets is not currently supported, because compileOnly dependencies must be present during the compilation of projects that depend on this project.

To ensure consistent compilation behaviour, compileOnly dependencies should be exposed as api dependencies.

Example:

    kotlin {
        sourceSets {
            nativeMain {
                dependencies {
                    compileOnly("org.example:lib:1.2.3")
                    // additionally add the compileOnly dependency as an api dependency:
                    api("org.example:lib:1.2.3")
                }
            }
        }
    }

This warning can be suppressed in gradle.properties:

    kotlin.suppressGradlePluginWarnings=IncorrectCompileOnlyDependencyWarning
```

mvicsokolova:
Fixes #448

### DIFF
--- a/integration-testing/examples/mpp-sample/build.gradle.kts
+++ b/integration-testing/examples/mpp-sample/build.gradle.kts
@@ -22,10 +22,16 @@ repositories {
 kotlin {
     jvm()
 
-    js()
+    js {
+        nodejs()
+    }
 
-    wasmJs {}
-    wasmWasi {}
+    wasmJs {
+        nodejs()
+    }
+    wasmWasi {
+        nodejs()
+    }
 
     macosArm64()
     macosX64()
--- a/integration-testing/examples/mpp-sample/build.gradle.kts
+++ b/integration-testing/examples/mpp-sample/build.gradle.kts
@@ -50,6 +50,11 @@ kotlin {
     }
 }
 
+// Workaround for KT-71203. Can be removed after https://github.com/Kotlin/kotlinx-atomicfu/issues/431
+atomicfu {
+    transformJs = false
+}
+
 tasks.withType<KotlinCompile<*>>().configureEach {
     kotlinOptions {
         freeCompilerArgs += listOf("-Xskip-prerelease-check")
--- a/integration-testing/examples/mpp-sample/build.gradle.kts
+++ b/integration-testing/examples/mpp-sample/build.gradle.kts
@@ -50,11 +50,6 @@ kotlin {
     }
 }
 
-// Workaround for KT-71203. Can be removed after https://github.com/Kotlin/kotlinx-atomicfu/issues/431
-atomicfu {
-    transformJs = false
-}
-
 tasks.withType<KotlinCompile<*>>().configureEach {
     kotlinOptions {
         freeCompilerArgs += listOf("-Xskip-prerelease-check")
--- a/integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
+++ b/integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
@@ -68,7 +68,7 @@ class MppProjectTest {
         mppSample.enableNativeIrTransformation = true
         mppSample.cleanAndBuild()
         // When Native IR transformations are applied, atomicfu-gradle-plugin still provides transitive atomicfu dependency
-        mppSample.mppNativeCheckAtomicfuInImplementation()
+        mppSample.mppNativeCheckAtomicfuInImplementation("macosX64")
         mppSample.checkConsumableDependencies()
         mppSample.mppCheckAtomicfuInApi("macosX64")
         // TODO: klib checks are skipped for now because of this problem KT-61143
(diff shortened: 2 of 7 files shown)


integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
```diff
@@ -33,28 +35,42 @@ class MppProjectTest {
...
     fun testMppNativeWithEnabledIrTransformation() {
         mppSample.enableNativeIrTransformation = true
         mppSample.cleanAndBuild()
         // When Native IR transformations are applied, atomicfu-gradle-plugin still provides transitive atomicfu dependency
-        mppSample.checkMppNativeImplementationDependencies()
+        mppSample.mppNativeCheckAtomicfuInImplementation()
+        mppSample.checkConsumableDependencies()
+        mppSample.mppCheckAtomicfuInApi("macosX64")
         // TODO: klib checks are skipped for now because of this problem KT-61143
         //mppSample.buildAndCheckNativeKlib()
```
fzhinkin, line 75: Can we enable it now?

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/framework/checker/DependenciesChecker.kt
```diff
@@ -24,52 +24,47 @@ private fun GradleBuild.checkAtomicfuDependencyIsAbsent(configurations: List<Str
...
-    checkAtomicfuDependencyIsPresent(listOf("macosX64CompileKlibraries"), commonAtomicfuDependency)
-    checkAtomicfuDependencyIsAbsent(listOf("macosX64MainImplementation"), commonAtomicfuDependency)
+internal fun GradleBuild.mppCheckAtomicfuInApi(targetName: String) {
+    checkAtomicfuDependencyIsPresent(listOf("${targetName}MainApi"), commonAtomicfuDependency)
 }
 
 // Checks Native target of an MPP project
-internal fun GradleBuild.checkMppNativeImplementationDependencies() {
-    checkAtomicfuDependencyIsPresent(listOf("macosX64CompileKlibraries", "macosX64MainImplementation"), commonAtomicfuDependency)
+internal fun GradleBuild.mppNativeCheckAtomicfuInImplementation() {
```
fzhinkin, line 66: Could you please pass the target name for the consistency with other checks?

Also, does it make sense to check other native targets?

integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/MppProjectTest.kt
mvicsokolova: Let's do this in a separate PR

integration-testing/examples/mpp-sample/build.gradle.kts
```diff
@@ -50,6 +50,11 @@ kotlin {
     }
 }
 
+// Workaround for KT-71203. Can be removed after https://github.com/Kotlin/kotlinx-atomicfu/issues/431
+atomicfu {
+    transformJs = false
```
fzhinkin, line 55: Should user apply the same change to their projects to use AFU w/ this PR applied without having "clashing `unique_name` property" error?

mvicsokolova: I'm preparing a fix for that rn: I'll set `transformJs` value to false by default

fzhinkin: But if, for some reason, they have this property explicitly enabled, they will get the cryptic error message anyway?

mvicsokolova: And I'll deprecate this property with a warning

fzhinkin: Maybe we should also print an explicit warning describing the problem to guide users toward the solution? Personally, I'm struggling to link `KLIB resolver blablabla unique_name blablabla` with `atomicfu.transformJs` :)

As of https://github.com/Kotlin/kotlinx-atomicfu/issues/431, I remember we agreed that JS transformation removal sounds OK, but the final decision was pending (some additional research over potential downsides was required).

mvicsokolova: Regarding #431 it's only related to JS IR, and I would do deprecation (and preliminary checks) of JS IR transformations for the next release.

`transformJs` only disabled/enabled legacy JS transformation and now, being still enabled by default, it just creates a new js directory, causing the problem, that you've mentioned. Removing `transformJs` flag and legacy transformation tasks would solve the problem 👀

fzhinkin: Got it, thanks.

> Removing transformJs flag and legacy transformation tasks would solve the problem 👀

Sure, but there are a few projects that are still using it.

mvicsokolova: I've prepared a PR removing legacy transformations and disabling/deprecating transformJs flag here: https://github.com/Kotlin/kotlinx-atomicfu/pull/478

With these changes, transformJs flag does not take any effect (specifically does not create an empty is directory which caused the above mentioned error), so for the user, who used transformJs there will just be a flag deprecation warning and the WA for this error will not be needed.

fzhinkin (approved):
I suggest releasing the changes as `0.26.0`. Otherwise, everything looks good to me.

PR #485, opened 2024-11-05
Remove findProperty to support ProjectIsolation

hfhbd:


atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -170,7 +170,7 @@ private fun KotlinVersion.atLeast(major: Int, minor: Int, patch: Int) =
 private fun Project.isCompilerPluginAvailable() = getKotlinVersion().atLeast(1, 6, 20)
 
 internal fun Project.getBooleanProperty(name: String) =
-    rootProject.findProperty(name)?.toString()?.toBooleanStrict() ?: false
+    providers.gradleProperty(name).orNull?.toBooleanStrict() ?: false
```
dkrasnoff, line 173: It is also worth looking into [ExtraPropertiesExtension](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.ExtraPropertiesExtension.html) extension

hfhbd: What exactly do you mean? AFAIK, you should avoid the (untyped) ext api too.

dkrasnoff: `findProperty` could previously look through extra properties. 
And I mean from the 'less breaking change' point of view we should support ext as well.

PR #498, opened 2024-12-19
Thread parking for Kotlin/Common

bbrockbernd:
Allows to pause and resume thread execution. On native platforms it is based on pthread_cond_wait, on JVM it uses LockSupport.

Threads can be pre unparked (calling unpark before park cancels the parking operation). And thread can be parker with a timeout.

Suspend the current thread by calling:
```Kotlin
ParkingSupport.park(timeout: kotlin.time.Duration)
// or
ParkingSupport.parkUntil(deadline: TimeMark)
// or park without timout
ParkingSupport.park(kotlin.time.Duration.INFINITE)
```

Resume a thread by calling:
```Kotlin
ParkingSupport.unpark(handle: ParkingHandle)
```

Get current thread reference by calling:
```Kotlin
val handle = ParkingSupport.currentThreadHandle()
```


atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/parking/KThread.kt
```diff
@@ -0,0 +1,19 @@
...
+
+actual class Parker actual private constructor() {
+    actual companion object {
+        actual fun park() = localKThread.get().parker.park()
+        actual fun parkNanos(nanos: Long) = localKThread.get().parker.parkNanos(nanos)
+        actual fun unpark(kThread: KThread) = localKThread.get().parker.unpark()
+    }
+}
+
+private val localKThread = ThreadLocal<KThread>().also { it.set(KThread()) }
```
dkhalanskyjb, line 18: This won't work, you need `ThreadLocal<KThread>.withInitialValue { KThread() }`. The reason: `.also { it.set(KThread()) }` will only be called once, on the thread that is the first to access `localKThread`, whereas the other threads will all have the default value, which is `null`.

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/parking/JvmParkingDelegator.kt
```diff
@@ -0,0 +1,43 @@
+package kotlinx.atomicfu.parking
+import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.locks.LockSupport
+import kotlin.time.DurationUnit
+import kotlin.time.TimeSource.Monotonic
+
+// Only for testing purposes
```
ndkoval, line 7: Not only for testing :)

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/parking/JvmParkingDelegator.kt
```diff
@@ -0,0 +1,43 @@
...
+internal class JvmParkingDelegator: ParkingDelegator {
+    private var thread: Thread? = null
+    private val atomicLong: AtomicLong = AtomicLong(0L)
```
ndkoval, line 10: Let's rename it to state, document that these states are "PARKED" and "UNPARKED", and use either `AtomicBoolean` provided AtomicFU or `AtomicIntegerFieldUpdater` if this is easier due to the compilation complexity.

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/parking/ParkingDelegator.kt
```diff
@@ -0,0 +1,13 @@
+package kotlinx.atomicfu.parking
+
+/**
+ * Internal utility that delegates the thread suspending and resuming calls calls in the platform specific way (darwin, linux, windows).
+ * On jvm delegates to LockSupport.Park.
+ */
+internal interface ParkingDelegator {
```
dkhalanskyjb, line 7: Unless I'm missing something, on each platform, there is a single implementation of this interface, so extracting it is premature, I feel.

```kotlin
internal expect class ParkingData

internal expect object ParkingDelegator {
    fun createRef(): ParkingData
    fun wait(ref: ParkingData) 
    fun timedWait(ref: ParkingData, nanos: Long)
    fun wake(ref: ParkingData)
    fun destroyRef(ref: ParkingData)
}
```
On the JVM, `internal actual typealias ParkingData = Unit` would achieve the same effect as what we have now.

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/parking/JvmParkingDelegator.kt
```diff
@@ -0,0 +1,43 @@
...
+    
+    override fun createRef(): Long {
+        thread = Thread.currentThread()
+        atomicLong.set(0L)
```
dkhalanskyjb, line 14: It seems to me like the atomic long here is a protection against spurious wake-ups. Is that it? I don't understand the utility of this safety mechanism. Code using thread parking is often already structured in a way that takes spurious wake-ups into account.

If we do remove the safety mechanism, however, it means that `JvmParkingDelegator` can be stateless and use `Thread` as the reference that's passed around. *That*, in turn, means that there's no need to create a parking delegator instance per thread parker.

atomicfu/src/androidNative64BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,72 @@
...
+        pthread_mutex_lock(ref.mut)
+        while (!ref.wake.value) {
+            pthread_cond_wait(ref.cond, ref.mut)
+        }
+        pthread_mutex_unlock(ref.mut)
+    }
+
+    actual override fun timedWait(ref: Any, nanos: Long) {
+        if (ref !is ParkingData) throw IllegalArgumentException("ParkingDelegator got incompatible parking object")
+        val ts = nativeHeap.alloc<timespec>().ptr
```
dkhalanskyjb, line 32: Minor: there's the https://kotlinlang.org/api/core/kotlin-stdlib/kotlinx.cinterop/mem-scoped.html function, which allows creating allocations more structurally.

```kotlin
memScoped {
  val ts = alloc<timespec>()
  clock_gettime(CLOCK_REALTIME.convert(), ts.ptr)
  // ...
  pthread_mutex_unlock(ref.mut)
} // `ts` gets deallocated when this lambda exits
```

atomicfu/src/jsAndWasmSharedMain/kotlin/kotlinx/atomicfu/parking/KThread.kt
```diff
@@ -0,0 +1,17 @@
...
+
+actual class Parker actual private constructor() {
+    actual companion object {
+        actual fun park() {}
+        actual fun parkNanos(nanos: Long) {}
+        actual fun unpark(kThread: KThread) {}
+    }
+}
+private val thisKthread = KThread()
+actual fun currentThreadId(): Long = 1
```
dkhalanskyjb, line 17: If we do decide to avoid spurious wake-ups, this implementation is incorrect. But in any case, I think it's also useless to provide it: I'm not sure what anyone using thread synchronization on single-threaded runtimes even expects. As part of the mutex implementation, this particular implementation makes sense, but not for other primitives.

Proposal: create a source set combining JVM and Native (in `kotlinx.coroutines`, we call it `concurrent`), and instead of providing parking in common code, only provide it in `concurrent`. Also, with this source set, it's easier to share tests between `jvmTest` and `nativeTest`: for most of them, it seems possible to put them in `concurrentTest`.

atomicfu/src/androidNative32BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,64 @@
...
+    
+    actual override fun timedWait(ref: Any, nanos: Long) {
+        if (ref !is ParkingData) throw IllegalArgumentException("ParkingDelegator got incompatible parking object")
+        val ts = nativeHeap.alloc<timespec>().ptr
+        ts.pointed.tv_sec = (nanos / 1_000_000).toInt()
+        ts.pointed.tv_nsec = (nanos % 1_000_000).toInt()
+        var rc = 0
+        pthread_mutex_lock(ref.mut)
+        while (!ref.wake.value && rc == 0) {
+            rc = pthread_cond_timedwait(ref.cond, ref.mut, ts)
```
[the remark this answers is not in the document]
bbrockbernd: No this should be added, thanks! I will try to run the tests on a 32Bit system.

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/parking/JvmParkingDelegator.kt
bbrockbernd: Moved all state to ParkingData similar to native

atomicfu/src/commonMain/kotlin/kotlinx/atomicfu/parking/ParkingDelegator.kt
bbrockbernd: During development I had multiple delegator types, (futex, posix, spinning) and this allowed me to quickly switch between them in tests. But this can now indeed be removed.

Done.

atomicfu/src/androidNative32BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,64 @@
...
+        }
+        pthread_mutex_unlock(ref.mut)
+        nativeHeap.free(ts)
+    }
+
+    actual override fun wake(ref: Any) {
+        if (ref !is ParkingData) throw IllegalArgumentException("ParkingDelegator got incompatible parking object")
+        pthread_mutex_lock(ref.mut)
+        ref.wake.value = true
```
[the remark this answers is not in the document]
bbrockbernd: In the case for native, the calling code should call the destroyRef function to deallocate. There can exists a race condition where:
- thread B (unparker) calls unpark and changes the state to free.
- thread A (parked) spuriously wakes up and since the state changed asumes it was woken up by a wake call. And continues to deallocate ParkingData.
- thread B continues to make the actual wake call, however the reference is destroyed.

So far I have not seen an easy way to solve this, and since parking threads is "expensive" to begin with I don't think it can hurt to leave it here? WDYT?

atomicfu/src/androidNative64BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
bbrockbernd: Done

atomicfu/src/androidNative32BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,64 @@
...
+        pthread_cond_signal(ref.cond)
+        pthread_mutex_unlock(ref.mut)
+    }
+
+
+    actual override fun destroyRef(ref: Any) {
+        if (ref !is ParkingData) throw IllegalArgumentException("ParkingDelegator got incompatible parking object")
+        pthread_mutex_destroy(ref.mut)
+        pthread_cond_destroy(ref.cond)
```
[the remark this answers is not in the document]
bbrockbernd: Done

atomicfu/src/jsAndWasmSharedMain/kotlin/kotlinx/atomicfu/parking/KThread.kt
bbrockbernd: Good suggestion! Added the sourceSet.
About the tests, spawning threads on Jvm and Native have different API's correct? Or would you suggest creating an expect/actual constructinon for thread manipulation?

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/parking/JvmParkingDelegator.kt
bbrockbernd: I have made the it stateless. However I do still check the spurious wakeups to have similar semantics to the native implementation. Why native needs this check has to do with memory deallocation (which is discussed further there)

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/parking/ThreadParker.kt
```diff
@@ -0,0 +1,82 @@
...
+                STATE_FREE -> {
+                    if (!state.compareAndSet(currentState, STATE_UNPARKED)) continue
+                    return
+                }
+
+                STATE_PARKED -> {
+                    if (!state.compareAndSet(currentState, STATE_FREE)) continue
+                    delegator.wake(atomicRef.value!!)
+                    return
+                }
```
dkhalanskyjb, line 74: Minor, style, subjective:
```suggestion
                STATE_FREE -> if (state.compareAndSet(currentState, STATE_UNPARKED)) return

                STATE_PARKED -> if (state.compareAndSet(currentState, STATE_FREE)) {
                    delegator.wake(atomicRef.value!!)
                    return
                }
```

atomicfu/src/androidNative32BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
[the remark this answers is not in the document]
dkhalanskyjb: Ok, I understand now that having this shared mutable state in some form is unavoidable, thanks! Yes, I agree that with the current implementation of `ThreadParker`, this change will lead to a data race. I think this can be avoided by reworking `ThreadParker` to only have one piece of atomic state (`atomic<FREE | UNPARKED | ParkingData>`), but I'm not sure if it would lead to anything useful.

atomicfu/src/jsAndWasmSharedMain/kotlin/kotlinx/atomicfu/parking/KThread.kt
dkhalanskyjb: We do `expect`/`actual` in our test source sets all the time, and it works without issues: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/concurrent/test/ConcurrentTestUtilities.common.kt.

atomicfu/src/androidNative32BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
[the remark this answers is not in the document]
bbrockbernd: Even with a single atomic state it seems difficult to me since we would still have to deallocate the native pthread_cond references. If we let the unparking thread deal with deallocation there exists a different race where the unpark call happens after the state change of park but before the actual `delegator.wait(...)` call.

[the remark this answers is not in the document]
dkhalanskyjb: I think this can be worked around by one more ABA-problem-resilient state (pseudocode below) if we do decide that it's worth doing:

```
fun park() {
  data = ParkingData
  if (state.cas(FREE, data)) {
    data.park()
    state.loop {
      when (it) {
        data -> if (state.cas(it, FREE)) {
          // unparker didn't see the data
          data.free()
          return
        }
        is UnparkAttempt -> if (state.cas(it, FREE)) {
          // we interruped the unparking attempt
          return
        }
        UNPARKED -> if (state.cas(UNPARKED, FREE)) {
          // unparker already left
          data.free()
          return
        }
        else -> error
      }
    }
  } else if (state.cas(UNPARKED, FREE)) {
    return
  } else {
    error
  }
}

fun unpark() {
  myUnparkAttempt = UnparkAttempt()
  state.loop {
    when (it) {
      is ParkingData -> if (state.cas(it, myUnparkAttempt)) {
        // we can attempt unparking. no one will deallocate the data in the meantime.
        it.unpark()
        if (!state.cas(myUnparkAttempt, UNPARKED)) {
          // the parker has already left, it's up to us to free the parking data
          it.free()
        }
        return
      }
      FREE -> if (state.cas(FREE, UNPARKED)) {
        // the parker didn't arrive yet
        return
      }
      UNPARKED -> return // someone already unparked this
      is UnparkAttempt -> return // someone is trying to unpark this
    }
  }
}
```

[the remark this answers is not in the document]
bbrockbernd: Hi, thanks for the suggestion! I have attempted to implement a version on your pseudocode. However, on native there still seems to be a race condition that leads to a thread being in the Unparked state while still waiting. While on JVM we can call `LockSupport.unpark()` before `LockSupport.park()`. On native signaling before waiting, leaves the thead waiting. 
Example:
- Thread P is parking, changes state Free -> data
- Thread U is unparking, observes state data, changes data -> unparkingAttempt, makes native singal call, changes unparkingAttempt -> Unparked
- Thread P makes native wait call.

From now any unpark call will have no effect since it is already in the unparked state.

Why it does work with spurious wakeup protection in native code, is beceause the "wake" state is observed and changed atomically with the signal and wake calls, protected by a `pthread_cond_mutex`.

Additionally, if there are two concurrent unpark calls while a thread is parked. It might result in the end state being "free". While one might expect the 2nd unpark call setting it in the "unparked" state.

Please, let me know if I mis understood something here!

atomicfu/src/jsAndWasmSharedMain/kotlin/kotlinx/atomicfu/parking/KThread.kt
bbrockbernd: Done

atomicfu/src/androidNative32BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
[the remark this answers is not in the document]
dkhalanskyjb: Can we enhance `data.park()` in my snippet so that it also accesses the state after acquiring the lock and [double-checks](https://en.wikipedia.org/wiki/Double-checked_locking) that it hasn't changed?

```kotlin
    data.park(doubleCheckBeforeParking = {
        state.value is ParkingData
    })
```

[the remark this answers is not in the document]
bbrockbernd: Than we would have to protect the state as well on the unparking side, since the state can still change even while holding the lock. Also, this wouldn't solve the concurrent unpark case (not sure if this is a big problem).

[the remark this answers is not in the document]
bbrockbernd: That seems to work. However, there is still the case where two threads unpark the parked thread concurrently and the final state is "free". While I would expect the state to be "unparked". Although, theoretically one could argue that you could observe it like both unpark calls were made before the park call, so not sure if this matters much..

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/parking/ThreadParker.kt
```diff
@@ -0,0 +1,158 @@
...
+private object Free : ParkingState
+
+// Parker is suspended and can be signaled with data.
+private class Parked(val data: ParkingData) : ParkingState
+
+// An unpark call happened while the parker was parked. In process of unparking.
+private class Unparking : ParkingState
+
+// An unpark call happened while unparking was already in process.
+private class UnparkedWhileUnparking(val attempt: Unparking) : ParkingState
```
dkhalanskyjb, line 157: What problem does this new state solve? Could you provide a non-linearizable interleaving that this state is protecting against?

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/parking/KThread.kt
```diff
@@ -0,0 +1,19 @@
...
+actual class Parker private actual constructor() {
+    actual companion object {
+        actual fun park(): Unit = localKThread.get().parker.park()
+        actual fun parkNanos(nanos: Long): Unit = localKThread.get().parker.parkNanos(nanos)
+        actual fun unpark(kThread: KThread): Unit = kThread.parker.unpark()
+    }
+}
+
+private val localKThread: ThreadLocal<KThread> = ThreadLocal.withInitial { KThread() }
+actual fun currentThreadId(): Long = Thread.currentThread().id
```
dkhalanskyjb, line 19:
```suggestion
internal actual fun currentThreadId(): Long = Thread.currentThread().id
```

atomicfu/build.gradle.kts
```diff
@@ -1,3 +1,4 @@
+import org.gradle.internal.impldep.org.junit.experimental.categories.Categories.CategoryFilter.exclude
```
dkhalanskyjb, line 1: This new import looks unused.

atomicfu/src/androidNative32BitMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,59 @@
...
+        callAndVerifyNative(0)  { pthread_mutex_unlock(ref.mut) }
+    }
+    
+    actual fun timedWait(ref: ParkingData, nanos: Long, shouldWait: () -> Boolean): Unit = memScoped {
+        val ts = alloc<timespec>().ptr
+
+        // Add nanos to current time
+        clock_gettime(CLOCK_REALTIME.toInt(), ts)
+        ts.pointed.tv_sec += nanos.toInt() / 1_000_000_000
+        ts.pointed.tv_nsec += nanos.toInt() % 1_000_000_000
```
dkhalanskyjb, line 31: Here's an incantation that compiles on both 32-bit and 64-bit targets:
```suggestion
        ts.pointed.tv_sec = (ts.pointed.tv_sec + nanos / 1_000_000_000).convert()
        ts.pointed.tv_nsec = (ts.pointed.tv_nsec + nanos % 1_000_000_000).convert()
```
This way, you can remove these custom `androidNative32BitMain`-like source sets and use the source sets provided out of the box like `androidNativeMain` (https://kotlinlang.org/docs/multiplatform-hierarchy.html#see-the-full-hierarchy-template)

atomicfu/src/androidNativeMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,56 @@
...
+        callAndVerifyNative(0) { pthread_mutex_unlock(ref.mut) }
+    }
+    
+    actual fun timedWait(ref: ParkingData, nanos: Long, shouldWait: () -> Boolean): Unit = memScoped {
+        val ts = alloc<timespec>().ptr
+
+        // Add nanos to current time
+        clock_gettime(CLOCK_REALTIME.toInt(), ts)
+        ts.pointed.tv_sec += nanos.toInt() / 1_000_000_000
+        ts.pointed.tv_nsec += nanos.toInt() % 1_000_000_000
```
dkhalanskyjb, line 31: There are a couple of important problems here.

First, an `Int` is about `-2.4 * 10^9 .. 2.4 * 10^9`. If that's nanoseconds, that's just 2.5 seconds: not nearly enough for practical applications of parking. So, converting `nanos` to an `Int` is certainly incorrect.

Second, even if we move `toInt` to outside the division/modulo, the problem is *still* there, just not as pronounced. `2^31` seconds is surely a lot, but `toInt()` doesn't always work like you'd expect. `println(((Long.MAX_VALUE / 3) / 1_000_000_000).toInt())` prints a large *negative* number. So, if someone were to pass `Long.MAX_VALUE / 3` to `timedWait` for some reason, they'd wait for a *negative* number of nanoseconds, which transforms this lock into a spinlock (whether or not we keep the spurious wakeup protection: if we do keep it, the spin lock is in the `Mutex` implementation, and if we don't, the user code will effectively implement the spin lock).

Third: all of this ignores the fact that `tv_sec + something` can *also* overflow, making `tv_sec` negative, again creating a spinlock.

Please ensure that in all cases where arithmetic overflow can happen, the mutex still behaves reasonably.

atomicfu/build.gradle.kts
```diff
@@ -128,52 +134,18 @@ kotlin {
...
-
+        
+        val concurrentMain by getting {}
+        val concurrentTest by getting {}
+        val nativeMain by getting { dependsOn(concurrentMain) }
+        val nativeTest by getting { dependsOn(concurrentTest) }
+        
+        val nativeUnixLikeMain by creating { dependsOn(nativeMain) }
+        val appleMain by getting { dependsOn(nativeUnixLikeMain) }
+        val linuxMain by getting { dependsOn(nativeUnixLikeMain) }
```
dkhalanskyjb, line 148: Minor: these lines can be moved to the `kotlin {` block where `concurrentMain` and `concurrentTest` are created, which makes a couple of lines here redundant and allows this whole `sourceSets` block to be removed.

dkhalanskyjb (changes requested):

atomicfu/src/nativeUnixLikeMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,57 @@
...
+        if (shouldWait()) callAndVerifyNative(0) { pthread_cond_wait(ref.cond, ref.mut) }
+        callAndVerifyNative(0) { pthread_mutex_unlock(ref.mut) }
+    }
+    
+    actual fun timedWait(ref: ParkingData, nanos: Long, shouldWait: () -> Boolean): Unit = memScoped {
+        val ts = alloc<timespec>().ptr
+
+        // Add nanos to current time
+        clock_gettime(CLOCK_REALTIME.convert(), ts)
+        ts.pointed.tv_sec = min(ts.pointed.tv_sec + nanos / 1_000_000_000, Int.MAX_VALUE.toLong()).convert()
```
dkhalanskyjb, line 31: The commit message says:

> Max waiting time is capped at [...] ~68 years.

An important clarification: it's not 68 years from now, it's 68 years from 1970, also known as the https://en.wikipedia.org/wiki/Year_2038_problem .

A solution that I imagine should work with both 32-bit and 64-bit `tv_sec`:

```kotlin
fun tvSecPlusNanos(tvSec: Long, nanos: Long): Long {
    val secondsToAdd = nanos / 1_000_000_000
    val newTvSec = tvSec + secondsToAdd
    // check for a change of sign in the result when the inputs have the same sign
    return if ((tvSec xor newTvSec) < 0 && tvSec >= 0) {
        // a very large Long number, and `Int.MAX_VALUE` when `toInt()/convert<Int>()` is called
        Long.MAX_VALUE - Int.MAX_VALUE - 1
    } else {
        newTvSec
    }
}

ts.pointed.tv_sec = tvSecPlusNanos(tvSec, nanos).convert()
```

I didn't test the function thoroughly, though, and I think it's worth doing, both with preset values (for smoke testing) and some randomly generated `tv_sec` (in the span of `-100..1000` years from now) and arbitrary positive `nanos` that validate the contract of this function.

I imagine the contract is:
* If clamping happens, it's because `tv_sec` is 32-bit, and the value is clamped to `Int.MAX_VALUE`. Nothing else we can do in this scenario, I suppose.
* If clamping doesn't happen, `(newTvSec - tvSec).seconds` should be close to `nanos.seconds`.

dkhalanskyjb (changes requested):

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/parking/JvmParkingDelegator.kt
```diff
@@ -0,0 +1,14 @@
...
+    
+    actual fun createRef(): ParkingData = ParkingData(Thread.currentThread())
+    actual fun wait(ref: ParkingData, shouldWait: () -> Boolean) = LockSupport.park()
+    actual fun timedWait(ref: ParkingData, nanos: Long, shouldWait: () -> Boolean) = LockSupport.parkNanos(nanos)
+    actual fun wake(ref: ParkingData) = LockSupport.unpark(ref.thread)
+    actual fun destroyRef(ref: ParkingData) {}
+    
+}
+
+internal actual class ParkingData(val thread: Thread)
```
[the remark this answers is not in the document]
bbrockbernd: It seems my comment was still "pending". As I tried to say, this solution fails on K1.
```
ACTUAL_WITHOUT_EXPECT] Actual typealias 'ParkingData' has no corresponding expected declaration The following declaration is incompatible because visibility is different:

- internal final expect class ParkingData defined in kotlinx. atomicfu. parking in file ParkingDelegator. kt
```

atomicfu/src/mingwMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,57 @@
...
+        if (shouldWait()) callAndVerifyNative(0)  { pthread_cond_wait(ref.cond, ref.mut) }
+        callAndVerifyNative(0)  { pthread_mutex_unlock(ref.mut) }
+    }
+
+    actual fun timedWait(ref: ParkingData, nanos: Long, shouldWait: () -> Boolean): Unit = memScoped {
+        val ts = alloc<timespec>().ptr
+
+        // Add nanos to current time
+        clock_gettime(CLOCK_REALTIME.toInt(), ts)
+        ts.pointed.tv_sec += nanos / 1_000_000_000
```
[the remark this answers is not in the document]
bbrockbernd: On windows `tv_sec` is Long and `tv_nsec` Int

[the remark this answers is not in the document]
dkhalanskyjb: Ok, fair. The max year Windows supports is 30827, according to https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-systemtime, so basically no system can be so misconfigured that an overflow is possible and still function correctly.

atomicfu/src/nativeUnixLikeMain/kotlin/kotlinx/atomicfu/parking/PosixParkingDelegator.kt
dkhalanskyjb: Oops. sorry, I just realized this function wasn't any better. I'll think some more.

bbrockbernd: For what I understood from the `convert()` function, is that it just calls `toInt()` and takes the least significant 32 bits. (Please correct if wrong). Therefore, I don't think it is possible to have one unified function that can return values larger than Int.Max (in case of 64Bit platform) _and_ maximum Int.max (in case of 32Bit platform).

dkhalanskyjb:
```kotlin
        ts.pointed.tv_sec = ts.pointed.tv_sec.addNanosToSeconds(nanos)

private fun Long.clampToInt(): Int = coerceIn(Int.MIN_VALUE.toLong(), Int.MAX_VALUE.toLong()).toInt() 
internal fun Int.addNanosToSeconds(nanos: Long): Int = (this + nanos / 1_000_000_000).clampToInt()
internal fun Long.addNanosToSeconds(nanos: Long): Long {
    val sum = this + nanos / 1_000_000_000
    return if ((this xor sum) < 0 && this >= 0) Long.MAX_VALUE else sum
}
```
Seems to do the trick!

dkhalanskyjb: A clarification, because it may not be obvious from my message: this if without creating any new source sets, because somehow, Kotlin manages to choose the correct overload for both 32-bit and 64-bit targets when I put this code in the source sets shared by them.

bbrockbernd: It'll be pretty far into the future before this overflows right? The maximum added value is `Long.MAX_VALUE / 1_000_000_000` which is around 300 years. And `Long.MAX_VALUE - 300 years` is still billions of years.

```kotlin
if ((this xor sum) < 0 && this >= 0) Long.MAX_VALUE else sum
```

dkhalanskyjb: Yes.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/parking/TimeArithmeticTests.kt
```diff
@@ -0,0 +1,47 @@
...
+                if (nanos < 0) assertTrue { updatedTime < currentTimeInSeconds }
+            }
+
+            // Test Int
+            repeat(1000) {
+                val currentTimeInInt = currentTimeInSeconds.toInt()
+                val nanos = Random.nextLong()
+                val updatedTime = currentTimeInInt.addNanosToSeconds(nanos)
+                if (nanos > 0) assertTrue { updatedTime > currentTimeInSeconds && updatedTime <= Int.MAX_VALUE }
+                if (nanos < 0) assertTrue { updatedTime < currentTimeInSeconds && updatedTime >= Int.MIN_VALUE }
```
dkhalanskyjb, line 41:
```suggestion
                if (nanos > 0) assertTrue { updatedTime > currentTimeInSeconds }
                if (nanos < 0) assertTrue { updatedTime < currentTimeInSeconds }
```
The `updatedTime <= Int.MAX_VALUE` and `updatedTime >= Int.MIN_VALUE` checks are supposed to be always `true` for an `Int`.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/parking/TimeArithmeticTests.kt
```diff
@@ -0,0 +1,47 @@
...
+                if (nanos > 0) assertTrue { updatedTime > currentTimeInSeconds }
+                if (nanos < 0) assertTrue { updatedTime < currentTimeInSeconds }
+            }
+
+            // Test Int
+            repeat(1000) {
+                val currentTimeInInt = currentTimeInSeconds.toInt()
+                val nanos = Random.nextLong()
+                val updatedTime = currentTimeInInt.addNanosToSeconds(nanos)
+                if (nanos > 0) assertTrue { updatedTime > currentTimeInSeconds && updatedTime <= Int.MAX_VALUE }
```
dkhalanskyjb, line 40: A stronger version of the check: if the value does not clamp, the `(updatedTime - currentTime) == nanos / 1_000_000_000`; if the value clamps, it's because as a `Long` value, it would exceed an `Int`. This describes what we want from `addNanosToSeconds` more precisely. As is, when I write `internal fun Long.addNanosToSeconds(nanos: Long): Long = this + nanos / 2_000_000_000`, for example, no tests fail.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/parking/TimedParkingTest.kt
```diff
@@ -0,0 +1,229 @@
...
+                throw t
+            }
+            finally { done.value = true }
+        }
+        thread = th
+    }
+    fun waitThrowing() {
+        thread!!.join()
+        throwIfError()
+    }
```
dkhalanskyjb, line 216: Minor, simplification:
```suggestion
internal class Fut(block: () -> Unit) {
    private val done = atomic(false)
    private val atomicError = atomic<Throwable?>(null)
    private val thread: TestThread = testThread {
        try { block() }
        catch (t: Throwable) {
            atomicError.value = t
            throw t
        }
        finally { done.value = true }
    }

    fun waitThrowing() {
        thread.join()
        throwIfError()
    }

```

atomicfu/build.gradle.kts
```diff
@@ -33,6 +59,12 @@ kotlin {
     wasmWasi {
         nodejs()
     }
+    
+    @Suppress("DEPRECATION") //https://github.com/Kotlin/kotlinx-atomicfu/issues/207
+    linuxArm32Hfp()
+
+    @OptIn(ExperimentalKotlinGradlePluginApi::class)
+    applyDefaultHierarchyTemplate()
```
dkhalanskyjb, line 67: An approach like https://github.com/Kotlin/kotlinx-datetime/blob/7d2764a54011365b8c6749f3107fa628cd167abc/core/build.gradle.kts#L41-L85 can help avoid manually specifying `concurrentMain` and such.

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,77 @@
...
+expect object ParkingSupport {
+    
+    /**
+     * Parks the current thread for [timout] duration.
+     * 
+     * Wakes up in the following cases:
+     * - A call to [ParkingSupport.unpark]
+     * - [timout] is exceeded
+     * - A spurious wakeup
+     * - (Only on JVM) The thread was interrupted
```
dkhalanskyjb, line 44: The documentation should mention that the *interrupted status* of the thread is untouched by the method (example: https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#interrupted--). It's also worth spelling out that without clearing the status manually, future attempts to `park` will automatically immediately resume, resulting in a spinlock, as this is quite a pitfall.

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,77 @@
...
+     * - [deadline] is exceeded
+     * - A spurious wakeup
+     * - (Only on JVM) The thread was interrupted
+     */
+    fun parkUntil(deadline: TimeMark)
+
+    /**
+     * Unparks the thread corresponding to [handle].
+     * If the corresponding thread was not parked the next park call will return immediately.
+     * A [ParkingHandle] can only be pre-unparked once.
```
dkhalanskyjb, line 62: Even as someone who knew the logic this documentation is describing, I still doubted myself: "Wait, are `ParkingHandle`s single-use?" I think it's worth explaining the idea more verbosely, like "only one call to `unpark` at a time can be buffered in this manner: even if several `unpark` calls are made before parking happens, the next parking process will not be affected and will suspend as expected", or something like that.

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,22 @@
...
+actual class ParkingHandle internal constructor() {
+    internal val parker: ThreadParker = ThreadParker()
+}
+
+actual object ParkingSupport {
+    actual fun park(timout: Duration) {
+        if (timout == Duration.INFINITE) threadLocalParkingHandle.parker.park()
+        else threadLocalParkingHandle.parker.parkNanos(timout.toLong(DurationUnit.NANOSECONDS))
+    }
+    actual fun parkUntil(deadline: TimeMark) = park(deadline.elapsedNow() * -1)
```
dkhalanskyjb, line 19: Question: can the code be simplified if we take `parkUntil` as the fundamental operation and then implement `parkNanos` on top of it? The idea is that POSIX exposes something that's similar to `parkUntil` already, so `waitWithDeadline` may be easier to write than `timedWait`, and `LockSupport.parkUntil` may be somewhat simpler than `LockSupport.parkNanos` (due to the simpler "is it time to exit yet?" check).

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,77 @@
...
+     */
+    fun unpark(handle: ParkingHandle)
+
+    /**
+     * Returns the [ParkingHandle] that can be used to unpark the current thread.
```
dkhalanskyjb, line 67:
```suggestion
     * Returns the [ParkingHandle] that can be used to [unpark] the current thread.
```
Easier to follow the breadcrumbs this way.

dkhalanskyjb (changes requested):
The implementation is almost ready to go, I believe. Good work! Only some documentation fixes are left, and there's maybe potential to simplify the code.

I haven't looked at the tests yet, though.

atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
bbrockbernd: AFAIK, a `TimeMark` is not directly convertible to nanos from epoch. The only way I have found so far is by somehow getting the current time from epoch and subtracting `TimeMark.elapsedNow` (which is converting to duration.)

Additionally, to get current time in nanos from epoch (on native) we either need `kotlinx-datetime` which doesn't seem to be more granular than milliseconds. Or rely on `clock_gettime` from POSIX and convert TimeMark to duration, which is what we currently do.

But I might be missing some api?

dkhalanskyjb: You're right, `TimeMark` does not expose its relationship to absolute time.

`kotlin.time.Instant` (which used to be in `kotlinx-datetime`) does provide nanosecond precision, and in fact, delegates to `clock_gettime`. So, with Kotlin 2.1.20, which introduced `Instant`, it's possible to do this (pseudocode):
```kotlin
fun parkUntil(deadline: TimeMark) {
  val estimatedTargetTime: Instant = Clock.System.now() - deadline.elapsedNow()
  sleepUntil(instant.epochSeconds, instant.nanosecondsOfSecond)
}
```

The problem is, we're not on Kotlin 2.1.20, so we indeed don't have access to `Instant`.

But! This gets me thinking: why do we need an `Instant` at all? It's semantically wrong. In fact, we may have a subtle bug.

See, `CLOCK_REALTIME` (https://man7.org/linux/man-pages/man2/clock_gettime.2.html), `kotlin.time.Instant`, and by default, `pthread_cond_timedwait` (https://docs.oracle.com/cd/E19120-01/open.solaris/816-5137/gfvid/index.html) all have to do with *system time*, that is, what your clock in the corner of the screen says. This clock may get out of sync, and then get back in sync when your system queries the Internet for the up-to-date time.

It seems like the following may happen:
* At 12:13:59, we park for three seconds. We are waiting until 12:14:02.
* One second later, at 12:14:00, we query the Internet and see that our clock is too fast and is into the future by two seconds.
* One second later, it's 12:13:59 again.
* Then, three seconds later, it's finally 12:14:02... but five seconds have passed.

`TimeSource.Monotonic`, though, works with elapsed time, which is what we want.

Proposed fix: https://stackoverflow.com/a/14397906. With this, we don't need to extract epoch time at all.

What do you think?

bbrockbernd: That is quite the edge case! However, the fix proposed relies on `pthread_condattr_setclock` which is not available on darwin. 

Also, I am wondering, is the kotlin `TimeSource.Monotonic` synced to the posix `CLOCK_MONOTONIC`?

dkhalanskyjb: > That is quite the edge case!

Indeed, but I'd estimate that every day, someone somewhere encounters this. Having clocks drift by a couple of seconds per day is typical.

> not available on darwin.

Oh, ok. If you also can't think of a way to fix this there, we need to at least document that on Native, the system clock is used and is susceptible to such issues.

> Also, I am wondering, is the kotlin `TimeSource.Monotonic` synced to the posix `CLOCK_MONOTONIC`?

Kotlin/Native delegates to C++'s `std::conditional<high_resolution_clock::is_steady, high_resolution_clock, steady_clock>`. I don't know how our C++ standard library implements those, but it's not necessarily a wrapper around POSIX.

atomicfu/build.gradle.kts
```diff
@@ -86,6 +90,7 @@ kotlin {
 
     sourceSets {
         commonMain.dependencies {
+            implementation("org.jetbrains.kotlinx:kotlinx-datetime:0.6.2")
```
dkhalanskyjb, line 93: This dependency is too heavy to be used here, and it also doesn't seem like it's used.

bbrockbernd: Oh my bad!

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/BarrierTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        val before = atomicArrayOfNulls<Int>(numberOfThreads)
+        init {repeat(numberOfThreads) {
+                after[it].value = 0
+                before[it].value = 0
+            }}
+    }
+    @Test
+    fun testBarrier() {
+        repeat(5) { iteration ->
+            println("Barrier test iteration $iteration")
```
dkhalanskyjb, line 22: This `println` doesn't seem useful, but will add extra noise in the logs.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/CyclicBarrierTest.kt
```diff
@@ -0,0 +1,148 @@
...
+            println("Stress test $iteration")
+            val threads = mutableListOf<Fut>()
+            val threadSetSize = (iteration + 1) * 5
+            val bar = CyclicBarrier(threadSetSize)
+            val syncBar = CyclicBarrier(threadSetSize * 5)
+            val ar = Arrs(threadSetSize * 5)
+            repeat(threadSetSize * 5) { tId ->
+                val t = Fut {
+                    repeat(50) { internalIteration ->
+                        sleepMills(Random.nextLong(100))
```
dkhalanskyjb, line 64: In total, these `sleepMillis` affect test execution time considerably. Are they necessary?

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/ThreadParkerTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        testThread {
+            sleepMills(1000)
+            ParkingSupport.unpark(mainThread)
+        }
+
+        val time = measureTime {
+            ParkingSupport.park(Duration.INFINITE)
+        }
+
+        assertTrue(time.inWholeMilliseconds > 900)
```
dkhalanskyjb, line 23: A more reliable (and still less time-consuming) way to test this would be to spin for several thousands of times and see if `park` ever finishes earlier than the corresponding `unpark` (basically, that happens-before is established between the code before `unpark` and the code after `park`).

Also, can't this test be flaky due to spurious wakeups?

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/ThreadParkerTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        val mainThread = ParkingSupport.currentThreadHandle()
+
+        testThread {
+            ParkingSupport.unpark(mainThread)
+        }
+
+        sleepMills(1000)
+        ParkingSupport.park(Duration.INFINITE)
+
+        assertTrue(true)
```
dkhalanskyjb, line 37: This line can be safely removed.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/TestThread.kt
```diff
@@ -0,0 +1,9 @@
+package kotlinx.atomicfu.locks
+
+internal fun testThread(doConcurrent: () ->  Unit): TestThread = TestThread { doConcurrent() }
```
dkhalanskyjb, line 3:
```suggestion
internal fun testThread(doConcurrent: () ->  Unit): TestThread = TestThread(doConcurrent)
```

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/ExchangerTest.kt
```diff
@@ -0,0 +1,57 @@
...
+internal class Exchanger<T> {
+    private val slot = atomic<Pair<ParkingHandle, T>?>(null)
+    fun exchange(item: T): T {
+        val myPair = Pair(ParkingSupport.currentThreadHandle(), item)
+        if (slot.compareAndSet(null, myPair)) {
+            while (slot.value == myPair) ParkingSupport.park(Duration.INFINITE)
+            val waiterPair = slot.value!!
+            slot.value = null
+            ParkingSupport.unpark(waiterPair.first)
+            return waiterPair.second!!
```
dkhalanskyjb, line 49:
```suggestion
            return waiterPair.second
```

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/ThreadParkerTest.kt
```diff
@@ -0,0 +1,57 @@
...
+            done = true
+        }
+        
+        // Allow thread to be parked before park call
+        sleepMillis(100)
+        ParkingSupport.unpark(parkingHandle!!)
+        
+        // Allow thread to wakeup
+        sleepMillis(400)
+        assertTrue(done)
```
dkhalanskyjb, line 28: Why not use `Fut` here instead of `testThread` and finish the test exactly when the thread stops? I suspect it takes much less than half a second to unpark a thread.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/ThreadParkerTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        
+        // allow testThread to start and publish handle
+        sleepMillis(100)
+
+        // Allow test 
+        ParkingSupport.unpark(parkingHandle!!)
+        
+        // Allow test thread to finish
+        sleepMillis(400)
+        assertTrue(done)
```
dkhalanskyjb, line 54: This delay seems preventable.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/ThreadParkerTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    fun unparkPark() {
+        var parkingHandle: ParkingHandle? = null
+        var done = false
+
+        testThread {
+            parkingHandle = ParkingSupport.currentThreadHandle()
+            
+            // Allow main thread to pre-unpark
+            sleepMillis(200)
```
dkhalanskyjb, line 41: This delay can also be avoided by introducing `@Volatile var unparked = false` that's set to `true` after `unpark`, and here, do `while (!unparked) { sleepMillis(10) }`.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/TimedParkingTest.kt
```diff
@@ -0,0 +1,230 @@
...
+        repeat(times) {
+            try {
+                return block()
+            } catch (t: Throwable) {
+                lastThrowable = t
+            }
+        }
+        val err = IllegalStateException("Failed after $times retries")
+        err.addSuppressed(lastThrowable?: IllegalStateException("Retry failed but no error"))
+        throw err
```
dkhalanskyjb, line 195: Minor, opinionated:
```suggestion
        lastThrowable?.let { throw IllegalStateException("Failed after $times retries").apply { addSuppressed(it) } }
```

dkhalanskyjb (changes requested):

fzhinkin:
There are small formatting issues here and there (like trailing spaces), please run the "Code/Reformat Code" in IJ (you can perform a bulk cleanup by selecting all the necessary files and then running the reformatting).

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/TimedParkingTest.kt
```diff
@@ -0,0 +1,231 @@
...
+import kotlin.test.assertTrue
+import kotlin.time.measureTime
+import kotlin.IllegalStateException
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.nanoseconds
+
+class TimedParkingTest {
+
+    @Test
+    fun testNanosFirstUnpark400() = retry(3) {
```
fzhinkin, line 14: Reliability of these tests worries me. 100ms slack feels huge, but in practice we can accidentally exceed it and got a test failure (and I guess we're already doing that given the test failures on Windows).

For deadline-specific scenarios, I don't think we need to unpark a future from the main thread at all, and we should only verify that it was parked for at least `timeout`-time-units.

For first-unpark-specific scenarios, we can't get away from verifying that the `park` was actually "interrupted" by the `unpark` call, but maybe we can increase the slack to a few minutes, instead of 100ms?

fzhinkin (changes requested):
Nicely done! I left a few comments regarding tests and documentation, but overall it looks good.

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ThreadParker.kt
```diff
@@ -0,0 +1,110 @@
...
+                    if (state.compareAndSet(myUnparkingState, Free)) return
+                    delegator.destroyRef(currentState.data)
+                    return
+                }
+            }
+        }
+    }
+}
+
+private interface ParkingState
```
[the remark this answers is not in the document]
bbrockbernd: Since a private interface cannot be extended outside the file I thought it wouldn't be necessary but it definitely can't hurt

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,84 @@
...
+     * - The given [deadline] has passed.
+     * - A spurious wakeup
+     * - (Only on JVM) The thread was interrupted. The interrupted flag stays set after wakeup.
+     * A future call to [park] this thread will return immediately, unless the `Thread.interrupted` flag is cleared.
+     */
+    fun parkUntil(deadline: TimeMark)
+
+    /**
+     * Unparks the thread corresponding to [handle].
+     * If [unpark] is called while the corresponding thread is not parked, the next [park] call will return immediately
```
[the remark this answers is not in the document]
bbrockbernd: This path is covered explicitly, in `ThreadParkerTest.unparkPark`. Additionally, the tests in `ThreadParkingStressTest` contain random sleeps before the park and unpark operations and should cover the pre-unpark path regularly.

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,84 @@
+package kotlinx.atomicfu.locks
+
+import kotlin.time.Duration
+import kotlin.time.TimeMark
+
+/**
+ * Parking and unparking support for threads on Kotlin/Native and Kotlin/JVM. 
+ * Can be used as a building block to create locks and other synchronization primitives.
+ *
+ * A call to [ParkingSupport.park] or [ParkingSupport.parkUntil] will suspend the current thread.
```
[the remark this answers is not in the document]
bbrockbernd: I see. My issue with "blocking" is that it could also refer to a busy wait and think it is nice to emphasize that the thread does not use any resources when it is parked. 

How do you like "pause"?

atomicfu/src/androidNativeMain/kotlin/kotlinx/atomicfu/locks/PosixParkingDelegator.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        callAndVerify(0) { pthread_condattr_destroy(attr) }
+        nativeHeap.free(attr)
+        return ParkingData(mut, cond)
+    }
+
+    actual inline fun wait(ref: ParkingData, shouldWait: () -> Boolean){
+        callAndVerify(0) { pthread_mutex_lock(ref.mut) }
+        if (shouldWait()) callAndVerify(0) { pthread_cond_wait(ref.cond, ref.mut) }
+        callAndVerify(0) { pthread_mutex_unlock(ref.mut) }
```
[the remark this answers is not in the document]
dkhalanskyjb: This is not an answer to your suggestion but a comment.
That would mean a bug somewhere in `pthread_cond_wait`, as it may not return a non-null value in this situation: https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_cond_wait.html . `callAndVerify` is an assertion, which should never actually fire; otherwise, the program can not meaningfully recover and we will will to fix the problem in the library.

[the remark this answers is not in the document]
bbrockbernd: Good one

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/TimedParkingTest.kt
bbrockbernd: Agreed, I tried to fix it with a few retries. But the tests on windows stay unreliable. Thanks for the suggestions!

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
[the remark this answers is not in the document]
fzhinkin: Sounds good to me

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,84 @@
...
+     * Parks the current thread until [deadline] is reached.
+     *
+     * Wakes up in the following cases:
+     * - A different thread calls [ParkingSupport.unpark].
+     * - The given [deadline] has passed.
+     * - A spurious wakeup
+     * - (Only on JVM) The thread was interrupted. The interrupted flag stays set after wakeup.
+     * A future call to [park] this thread will return immediately, unless the `Thread.interrupted` flag is cleared.
+     */
+    fun parkUntil(deadline: TimeMark)
```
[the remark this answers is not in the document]
bbrockbernd: Added them to the timed tests

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,84 @@
...
+ *   ParkingSupport.park(Duration.INFINITE)
+ * }
+ * ```
+ * 
+ * Example usage unparker thread:
+ * ```Kotlin
+ * state.value = WAKE
+ * ParkingSupport.unpark(handleReference.value)
+ * ```
+ */
```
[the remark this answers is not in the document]
bbrockbernd: What OptIn would you suggest, `ExperimentalMultiplatform` ?

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/CyclicBarrierTest.kt
```diff
@@ -1,3 +1,5 @@
+@file:OptIn(ExperimentalThreadBlockingApi::class)
```
dkhalanskyjb, line 1: I think the opt-in should be put into `buildSrc/src/main/kotlin/kotlin-base-conventions.gradle.kts` (like `ExperimentalForeignApi` already is). Then, it should be possible to avoid adding `OptIn` everywhere throughout the library.

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Returns the [ParkingHandle] corresponding to the current thread.
+     * This [ParkingHandle] should be shared with other threads which allow them to [unpark] the current thread.
+     *
+     * A [ParkingHandle] is uniquely associated with a specific thread, maintaining a one-to-one correspondence.
+     * When the _same_ thread makes multiple calls to [currentThreadHandle],
+     * it always returns the _same_ [ParkingHandle].
+     *
+     * Note: as this function returns a unique [ParkingHandle] for each thread it should not be cached or memoized.
```
fzhinkin, line 89: But the example illustrating how to use parking caches (well, publishes, but its hard to tell the difference) the handle:
```
 * // publish my parking handle
 * handleReference.value = ParkingSupport.currentThreadHandle()
```

fzhinkin (commented):
There are a few nits, but otherwise the change looks good.

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
```diff
@@ -0,0 +1,122 @@
...
+     * Parks the current thread for [timeout] duration.
+     *
+     * Wakes up in the following cases:
+     * - A different thread calls [ParkingSupport.unpark].
+     * - The [timeout] is exceeded.
+     * - A spurious wakeup
+     * - (Only on JVM) The thread was interrupted. The interrupted flag stays set after wakeup.
+     * A future call to [park] this thread will return immediately, unless the `Thread.interrupted` flag is cleared.
+     */
+    fun park(timeout: Duration)
```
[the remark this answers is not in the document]
bbrockbernd: Currently a negative waiting time already returns immediately. Which is in IMO an expected outcome. 

Additionally, when the users logic calculates the _time left to wait_, the outcome could be negative as some internal deadline might have passed during this calculation. Making negative numbers invalid would force the user to do additional error handling or input validation.

While currently, the user already has to re-evaluate the conditions after every wake-up due to the chance for spurious wake-ups.

So I think it would make use of this API more cumbersome and doesn't give much benefit. 

WDYT?

atomicfu/src/jvmMain/kotlin/kotlinx/atomicfu/locks/JvmParkingDelegator.kt
```diff
@@ -0,0 +1,15 @@
...
+
+    actual fun createRef(): ParkingData = Thread.currentThread()
+    actual fun wait(ref: ParkingData, shouldWait: () -> Boolean) = LockSupport.park()
+    actual fun timedWait(ref: ParkingData, nanos: Long, shouldWait: () -> Boolean) = LockSupport.parkNanos(nanos)
+    actual fun wake(ref: ParkingData) = LockSupport.unpark(ref)
+    actual fun destroyRef(ref: ParkingData) {}
+
+}
+
+internal actual typealias ParkingData = Thread
```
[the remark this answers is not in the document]
bbrockbernd: Done

atomicfu/src/concurrentMain/kotlin/kotlinx/atomicfu/locks/ParkingSupport.kt
bbrockbernd: I've changed it. Hope it is clearer this way. 

Otherwise, we could maybe remove this note all together since we already mentioned there is a one-to-one correspondance between a thread and the handle?

fzhinkin: It's much better now, thanks!

And I think it's better to leave the note as it might help some users.

dkhalanskyjb (approved):
Ok, what next? @ndkoval, do you also want to review this, or can we merge the PR?

ndkoval:
@dkhalanskyjb please merge the PR.

PR #499, opened 2024-12-19
Apple QoS support via donation

stefanhaustein:
Addressing issue #462


stefanhaustein:
Platform dependencies change diagram: https://docs.google.com/presentation/d/1JtQ5GsjDmhtPKCYkzxPwRp-fD58NiKMw4nIi85_M7kg/edit?usp=sharing

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -0,0 +1,297 @@
...
+// from the atomicfu implementation.
+public actual open class SynchronizedObject {
+
+    companion object {
+        private const val NO_OWNER = 0L
+    }
+
+    private val owner: AtomicLong = atomic(NO_OWNER)
+    private var reEnterCount: Int = 0
+    private val waiters: AtomicInt = atomic(0)
```
qwwdfsad, line 67: Calling this `waiters` is semantically confusing: if there exists only a single thread that locks the lock, `waiters` fluctuates between 0 and 1.

Have you considered renaming it? (e.g. `threadsOnLock` or something like that) and dropping a comment?

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -0,0 +1,297 @@
...
+
+// Based on the compose-multiplatform-core implementation with added qos and the pool back-ported
+// from the atomicfu implementation.
+public actual open class SynchronizedObject {
+
+    companion object {
+        private const val NO_OWNER = 0L
+    }
+
+    private val owner: AtomicLong = atomic(NO_OWNER)
```
qwwdfsad, line 65: Please consider naming it `ownerThreadId` -- generally `owner` means `Any` in our lingua

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -0,0 +1,297 @@
...
+                // There is an existing override, but we need to go higher.
+                if (ourQosClass > qosOverrideQosClass) {
+                    pthread_override_qos_class_end_np(qosOverride)
+                    qosOverride = pthread_override_qos_class_start_np(lockOwner.toCPointer(), qos_class_self(), 0)
+                    qosOverrideQosClass = ourQosClass
+                }
+            } else {
+                // No existing override, check if we need to set one up.
+                memScoped {
+                    val lockOwnerQosClass = alloc<UIntVar>()
```
qwwdfsad, line 169: It's pretty much likely that this allocation will be happen on each contention and it's a real `malloc`.

Is it possible to extract this `alloc` into `DonatingMonitor` level and free it within cleaner?

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -0,0 +1,297 @@
...
+                    val lockOwnerRelPrio = alloc<IntVar>()
+                    pthread_get_qos_class_np(lockOwner.toCPointer(), lockOwnerQosClass.ptr, lockOwnerRelPrio.ptr)
+                    if (ourQosClass > lockOwnerQosClass.value) {
+                        qosOverride = pthread_override_qos_class_start_np(lockOwner.toCPointer(), qos_class_self(), 0)
```
qwwdfsad, line 173: We can reuse `ourQosClass` here instead of calling `qos_class_self`, can't we?

qwwdfsad (changes requested):
Great job!

I have a few essential comments on Apple-related implementation, and it's good to go. To avoid extra-round trips please take a look at new warnings produced by `./gradlew build`.

I would also kindly ask (but I'll accept the CR without it) to align non-apple `SynchronizedObject` with the current implementation as well -- you can copy-paste it but remove QoS mentions, or you can make this code part of `nativeMain` with `expect` `DonatingMonitor`, whatever is easier. Having two **different** implementations going to be real PITA here

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -0,0 +1,297 @@
...
+        return try {
+            block()
+        } finally {
+            monitor.nativeMutex.exit()
+        }
+    }
+
+    @OptIn(kotlin.experimental.ExperimentalNativeApi::class)
+    private class DonatingMonitor {
+        val nativeMutex = mutexPool.allocate()
```
bbrockbernd, line 142: Since mutexPool.allocate() is only called once and therefore only a single node is used. I believe we can get rid of MutexPool all together. By replacing this line for `val nativeMutex = NativeMutexNode()`. And replacing the cleaner with `val cleaner = createCleaner(nativeMutex) { it.dispose() }`.

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
bbrockbernd, line 173: I think there is a race condition possible here:
Let’s say we have thread 1 and 2 with low priority and thread 3 with high priority. 

Thread 1 is holding the lock and doing work, thread 2 is suspended and waiting for the lock to be released. There is no qos override since 1 and 2 have equal priority. 

Now, if thread 3 (high priority) tries to acquire the lock it would reach this point (line 173) in the code (please correct me if I’m wrong). If in the meantime thread 1 releases the lock and thread 2 therefore acquires it, then the `lockOwner` reference is outdated and we would override priority from thread 1 and not the new lock holder thread 2.

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -0,0 +1,297 @@
...
+
+@ThreadLocal
+var currentThreadId = 0L
+
+// Based on the compose-multiplatform-core implementation with added qos and the pool back-ported
+// from the atomicfu implementation.
+public actual open class SynchronizedObject {
+
+    companion object {
+        private const val NO_OWNER = 0L
```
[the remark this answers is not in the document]
stefanhaustein: Done

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
```diff
@@ -0,0 +1,297 @@
...
+            donateQos(lockOwner)
+            nativeMutex.wait()
+            clearDonation()
+        }
+
+        private fun donateQos(lockOwner: Long) {
+            if (lockOwner == NO_OWNER) {
+                return
+            }
+            val ourQosClass = qos_class_self() as UInt
```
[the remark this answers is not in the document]
stefanhaustein: Removed

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
stefanhaustein: Renamed as suggested

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
stefanhaustein: renamed as suggested

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
stefanhaustein: Shouldn't this just be a stack allocation (via memScoped)? It's just used short-lived for getting the values back, perhaps I am holding it wrong?

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
stefanhaustein: Done

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
stefanhaustein: While reducing the number of classes and objects by one, I think this might reduce readability. I don't have a strong option either way, but I'd prefer to keep the pool aligned with the implemenations for the other platforms, at least until we have merged them in a followup?

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
stefanhaustein: Thread 2 shouldn't be able to acquire the mutex in waitForUnlockAndLock while Thread 3 holds it on line 173?

stefanhaustein:
> I have a few essential comments on Apple-related implementation, and it's good to go. To avoid extra-round trips please take a look at new warnings produced by `./gradlew build`.

I checked the IDE for warnings and fixed the threadlocal package. Not sure about the redundant "public" modifiers? For some bits (forked from the compose implementation) I was not sure to which extent they were outdated or intended for some reason. Unfortunately, `gradlew build` is full of gazillions of expect / actual warnins. I did a quick search for the flag in combination with gradlew but did not see anything obvious....

> I would also kindly ask (but I'll accept the CR without it) to align non-apple `SynchronizedObject` with the current implementation as well -- you can copy-paste it but remove QoS mentions, or you can make this code part of `nativeMain` with `expect` `DonatingMonitor`, whatever is easier. Having two **different** implementations going to be real PITA here

SGTM but let me do this in a followup.

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
bbrockbernd: Ah I see, donation happens between pthread_mutex_lock and pthread_mutex_wait. My bad!

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
bbrockbernd: I would argue that having unnecessary code decreases readability, but as you said this will probably be resolved when the apple and non-apple implementations are unified.

stefanhaustein:
Does this make sense? Not sure why the requested change is still highlighted -- it should be addressed in the updated branch....

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
qwwdfsad: `memScoped` is still a real allocation, we have no stack allocation in K/N AFAIR

qwwdfsad:
I've run the remote build. In the meantime, PTAL at the malloc commentary

qwwdfsad (commented):
.

atomicfu/src/appleMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt
stefanhaustein: Done

stefanhaustein:
> I've run the remote build. In the meantime, PTAL at the malloc commentary

Done.

qwwdfsad:
Functional tests passed, now I'm on benchmarks. Almost there :)

qwwdfsad (approved):
All good, thanks!

Could you please consider the follow-up that removed the duplicate implementation with a different mechanism (non-apple locks)?

PR #504, opened 2025-01-23
Fix jvm target validation error

Tapchicoma:
Upcoming changes in Kotlin Gradle Plugin introduce JavaCompile tasks to the KMP/JVM target. This will lead to JVM target validation error when Kotlin JVM compile task will have different JVM target from JavaCompile task.
    
  To avoid this - switched configuring JVM target to the JVM toolchain configuration, which also configures JavaCompile tasks.

### DIFF
--- a/atomicfu/build.gradle.kts
+++ b/atomicfu/build.gradle.kts
@@ -11,6 +11,7 @@ plugins {
 }
 
 kotlin {
+    jvmToolchain(8)
 
     // JS -- always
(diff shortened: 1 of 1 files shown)


mvicsokolova (approved):

mvicsokolova:
Thank you!

Tapchicoma:
@tbogdanova could also pick this change into community branch?

PR #507, opened 2025-02-04
Set JDK toolchain version to 11 for transform tasks

mvicsokolova:


atomicfu/build.gradle.kts
```diff
@@ -210,7 +210,7 @@ val transformFU by tasks.registering(JavaExec::class) {
     classpath(transformer)
     inputs.dir(classesPreAtomicFuDir)
     outputs.dir(classesPostTransformFU)
-    launcherForJdk(9)
+    launcherForJdk(11)
```
fzhinkin, line 213: Does it make sense to extract launcher JDK version into a constant (or a property) so that the next time we'll need to change it in a single place?

mvicsokolova: Sure, I've pushed a fixup

fzhinkin (approved):

PR #509, opened 2025-02-10
Build warnings as errors

Tapchicoma:
I've enabled warning-as-errors mode in build scripts and Gradle plugin compilation to avoid usages of deprecated methods.

### DIFF
--- a/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
+++ b/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
@@ -15,7 +15,6 @@ import org.gradle.api.tasks.testing.*
 import org.gradle.jvm.tasks.*
 import org.gradle.util.*
 import org.jetbrains.kotlin.gradle.dsl.*
-import org.jetbrains.kotlin.gradle.dsl.KotlinCompile
 import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.tasks.*
 import java.io.*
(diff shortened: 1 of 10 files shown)


atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt
```diff
@@ -267,7 +267,7 @@ private fun Project.withKotlinTargets(fn: (KotlinTarget) -> Unit) {
     }
 }
 
-private fun KotlinCompile<*>.setFriendPaths(friendPathsFileCollection: FileCollection) {
+private fun KotlinCompilationTask<*>.setFriendPaths(friendPathsFileCollection: FileCollection) {
```
Tapchicoma, line 270: I am not sure `KotlinCompilationTask` type is available in KGP 1.7.0. In general, I think it is time to bump minimal supported versions of Gradle and KGP in this project.

mvicsokolova: AFUPlugin requires Kotlin plugin version to be at least `1.9.0` already, so I guess, it's ok? 


It was hardcoded [here](https://github.com/Kotlin/kotlinx-atomicfu/blob/ff6051133d5b9509317ef5e1e7016e6f465b72b3/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt#L76), instead of using the declared const value.

Tapchicoma: With Kotlin `1.9.0` it should fine - this interface was introduced in Kotlin `1.8.0` alongside `compilerOptions` DSL.

Tapchicoma: I could try to update the logic to use the constant as I was also confused with it value. Are you ok with it?

mvicsokolova (approved):

PR #510, opened 2025-02-10
Fix deprecation warnings coming from KGP

ALikhachev:
I'm updating deprecation levels in the scope of https://youtrack.jetbrains.com/issue/KT-68597, and it causes the compilation to fail because deprecated symbols are used in the plugin


ALikhachev:
Closed in favour of https://github.com/Kotlin/kotlinx-atomicfu/pull/509

PR #524, opened 2025-03-01
Use SPDX identifier in POMs

TheMrMilchmann:
This replaces the custom name with an SPDX identifier to enable tooling to automatically detect the correct license. Using an SPDX identifier is recommended [by the official Maven documentation](https://maven.apache.org/pom.html#Licenses).

See https://spdx.org/licenses/Apache-2.0.html


fzhinkin (approved):
@TheMrMilchmann, thank you for the contribution!

PR #531, opened 2025-04-28
Implemented additional compiler flags for KT-75078

mvicsokolova:


mvicsokolova:
@shanshin hi! Could you please have a look, if my changes for compiler flags make sense?

https://youtrack.jetbrains.com/issue/KT-75078

buildSrc/src/main/kotlin/KotlinConfiguration.kt
```diff
@@ -83,3 +82,49 @@ fun getOverridingKotlinApiVersion(project: Project): String? {
...
+    }
+}
+
+/**
+ * Pass additional CLI options to the Kotlin compiler.
+ */
+fun KotlinCommonCompilerOptions.addExtraCompilerFlags(project: Project) {
+    val extraOptions = project.rootProject.properties["kotlin_additional_cli_options"] as? String
+    if (extraOptions != null) {
+        LOGGER.info("""Adding extra compiler flags '$extraOptions' for a compilation in the project $${project.name}""")
```
fzhinkin, line 125: It makes sense to use the same logging format across all our projects, and the one from `kotlinx-serialization` looks fine: https://github.com/Kotlin/kotlinx.serialization/pull/2996/files#diff-09e3a69c8143d7af22e62f518f556102bbdea00485f9e30be972303bc6c7dc19R58

mvicsokolova: Thank you) I've fixed that

buildSrc/src/main/kotlin/KotlinConfiguration.kt
```diff
@@ -83,3 +82,48 @@ fun getOverridingKotlinApiVersion(project: Project): String? {
...
+        "-Xrender-internal-diagnostic-names",
+    )
+}
+
+/**
+ * Use `kotlin_Werror_override` property to override the state of -Werror:
+ * `true` means that warnings should be treated as errors,`false` means that they should not.
+ */
+private fun warningsAreErrorsOverride(project: Project): Boolean? =
+    when (val prop = project.rootProject.properties["kotlin_Werror_override"] as? String) {
```
shanshin, line 101: It's better to use `providers.gradleProperty` instead of `project.rootProject.properties`, see [the changes](https://github.com/Kotlin/kotlinx.serialization/pull/2997/files#diff-09e3a69c8143d7af22e62f518f556102bbdea00485f9e30be972303bc6c7dc19R8) in the serialization

buildSrc/src/main/kotlin/KotlinConfiguration.kt
```diff
@@ -83,3 +82,48 @@ fun getOverridingKotlinApiVersion(project: Project): String? {
...
+    } else {
+        allWarningsAsErrors.set(true)
+    }
+}
+
+/**
+ * Pass additional CLI options to the Kotlin compiler.
+ */
+fun KotlinCommonCompilerOptions.addExtraCompilerFlags(project: Project) {
+    val extraOptions = project.rootProject.properties["kotlin_additional_cli_options"] as? String
```
shanshin, line 123: Same about `project.rootProject.properties`

buildSrc/src/main/kotlin/KotlinConfiguration.kt
mvicsokolova: Fixed that

woainikk (approved):

mvicsokolova:
@fzhinkin, hi!) Are we good to merge these changes?

buildSrc/src/main/kotlin/KotlinConfiguration.kt
```diff
@@ -83,3 +83,49 @@ fun getOverridingKotlinApiVersion(project: Project): String? {
...
+        allWarningsAsErrors.set(true)
+    }
+}
+
+/**
+ * Pass additional CLI options to the Kotlin compiler.
+ */
+fun KotlinCommonCompilerOptions.addExtraCompilerFlags(project: Project) {
+    project.providers.gradleProperty("kotlin_additional_cli_options").orNull?.let { options ->
+        options.removeSurrounding("\"").split(" ").filter { it.isNotBlank() }.forEach {
```
fzhinkin, line 127: We discussed it with @woainikk, and it seems like `removeSurrounding("\"")` is unnecessary here.

woainikk: after implementing it in serialization, we understand that we can avoid it by passing parameters in other way, so yes, it's not needed anymore, sorry for confusing

PR #536, opened 2025-06-03
Speedup integration tests

fzhinkin:
Speedup integration tests execution by:
- abstaining from running redundant commands;
- combining multiple commands into a single Gradle invocation, where possible;
- enable parallel tests suite execution;
- split the slowest suite (`MppProjectTest`) into a several suite to exploit parallel suite execution.

As a result, on Windows (which is the slowest platform w.r.t. tests execution), execution times improves by 15 minutes.


mvicsokolova:
Looks good, thank you 🙏

PR #540, opened 2025-06-16
KT-78372 Disable Klibs cross compilation

AYastrebov:
Since Kotlin 2.2.20 we’ve been planning to enable [Klibs cross compilation by default](https://youtrack.jetbrains.com/issue/KT-76421/Stabilize-klib-cross-compilation-on-different-platforms). For atomicfu we now need to explicitly disable cross compilation `kotlin.native.enableKlibsCrossCompilation=false`

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,7 +6,7 @@ version=0.28.0-SNAPSHOT
 group=org.jetbrains.kotlinx
 
 kotlin.native.ignoreDisabledTargets=true
-
+kotlin.native.enableKlibsCrossCompilation=false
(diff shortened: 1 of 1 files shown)


AYastrebov:
> @AYastrebov could you please elaborate why that's necessary? Are there any negative consequences of cross-compiling AFU's native libs?

Cross compilation with cinterops is only possible on supported host systems. If we find cinterops during the configuration phase, then the ERROR diagnostic will be fired and build will be cancelled.

woainikk (approved):

PR #548, opened 2025-07-03
Remove usage of internal Gradle API

adam-enko:
Gradle tasks should be subtypes of DefaultTask, not the internal `ConventionTask`


mvicsokolova:
I'm sorry, I cannot recall at all the details, why I used `DefaultTask` back then :(

fzhinkin (approved):

PR #552, opened 2025-07-08
Fix test name to use ASCII characters only.

chaoren:


fzhinkin (approved):
@chaoren thanks for noticing it and fixing!

PR #555, opened 2025-07-22
Update `kotlin-for-gradle-plugin` to `2.2`

lunakoly:
This is required for 2.3 bootstrapping as bumping the version to 2.3 produces artifacts with metadata 2.3, and KGP brings compiler `2.0.21` which can't read it.

^KT-76136


fzhinkin:
@lunakoly but where `2.0.21` is coming from? The plugin is currently built w/ `2.1.0`, not `2.0.21`

PR #577, opened 2026-02-06
Fix CustomCountDownLatch

murfel:
Following up on #498. We need to use a multiplatform CountDownLatch in https://github.com/Kotlin/kotlinx.coroutines/pull/4606, so let's fix it here first and then I'll merrily copy-paste it into kx.coroutines.

### DIFF
--- a/atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/LatchTest.kt
+++ b/atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/LatchTest.kt
@@ -82,8 +82,8 @@ class CustomCountDownLatch(count: Int) {
 }
 
 private class MPSCQueueLatch<E> {
-    private val head = atomic(Node<E>(null))
-    private val tail = atomic<Node<E>?>(head.value) // if null, then closed
+    private val head = Node<E>(null)
+    private val tail = atomic<Node<E>?>(head) // if null, then closed
 
     fun enqueue(element: E): Boolean {
         val node = Node(element)
@@ -100,12 +100,12 @@ private class MPSCQueueLatch<E> {
 
     fun drain(action: (E) -> Unit) {
         close()
-        var node = head.value.next.value
+        var node = head.next.value
         while (node != null) {
             action(node.element!!)
             node = node.next.value
         }
-        head.value.next.value = null
+        head.next.value = null
     }
 
     private fun close() {


atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/LatchTest.kt
```diff
@@ -63,53 +64,59 @@ class LatchTest {
...
-                tail.compareAndSet(curTail, node)
-                return
-            } else tail.compareAndSet(curTail, curTail.next.value!!)
+    private val tail = atomic<Any>(head.value)
+
+    fun enqueue(element: E): Boolean {
+        val node = Node(element)
+        tail.loop {
+            if (it === finished) return false
+            if ((it as Node<E>).next.compareAndSet(null, node)) {
```
dkhalanskyjb: Looks like `tail` can only contain `Node<E>` or `null`. Wouldn't `val tail = atomic<Node<E>?>(...)` do the trick just as well, with `finished == null`? This would help not to annoy the compiler.

If not, better to at least add a suppression to the unchecked cast so that warnings don't pollute logs.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/LatchTest.kt
```diff
@@ -63,53 +64,59 @@ class LatchTest {
...
 class CustomCountDownLatch(count: Int) {
     private val c = atomic(count)
-    private val waiters = MSQueueLatch<ParkingHandle>()
+    private val waiters = MPSCQueueLatch<ParkingHandle>()
 
     fun await() {
         val thread = ParkingSupport.currentThreadHandle()
-        waiters.enqueue(thread)
-        while (c.value > 0) ParkingSupport.park(Duration.INFINITE)
+        if (waiters.enqueue(thread)) {
```
dkhalanskyjb: Worth checking `c.value` before the `enqueue` call as well, to avoid unnecessarily allocating a `Node` when the latch is already lowered.

atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/LatchTest.kt
```diff
@@ -63,53 +64,59 @@ class LatchTest {
...
-                currentHeadNext.element = null
-                return element
-            }
+    fun drain(action: (E) -> Unit) {
+        close()
+        var node = head.value.next.value
+        while (node != null) {
+            action(node.element!!)
+            node = node.next.value
+        }
```
dkhalanskyjb: There's a memory leak here: as long as the `CountDownLatch` lives, the full contents of the queue will be reachable. The procedure of removing values from the `head` and setting `element` of the new `head` to `null`, as is done in `dequeue`, is still valuable.

PR #581, opened 2026-02-11
Add logic to handle changed `.kotlin_module` file name in test

Tapchicoma:
Due to KT-69701, Kotlin 2.4.0 changes `.kotlin_module` file name to align with other platforms. This change requires adjusting the test logic.

### DIFF
--- a/integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/JvmProjectTest.kt
+++ b/integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/JvmProjectTest.kt
@@ -55,14 +55,25 @@ class JvmProjectTest {
 
         jvmSample.build().apply {
             assertTrue { buildClassesAtomicfuDir.exists() }
-            assertEquals(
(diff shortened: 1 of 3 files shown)


integration-testing/src/functionalTest/kotlin/kotlinx.atomicfu.gradle.plugin.test/cases/JvmProjectTest.kt
```diff
@@ -58,10 +58,10 @@ class JvmProjectTest {
             assertEquals(
                 """
                 build/classes/atomicfu/main/IntArithmetic.class
-                build/classes/atomicfu/main/META-INF/jvm-sample.kotlin_module
+                build/classes/atomicfu/main/META-INF/kotlinx.atomicfu.examples_jvm-sample.kotlin_module
```
fzhinkin, line 61: @Tapchicoma, the test needs the same version-dependent behavior as the other test from this PR. Right now, it passes w/ a Kotlin built from the master, but fails w/ the Kotlin version used in the the project by default: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxAtomicfu_BuildAggregatedLinuxWindows/5999752?buildTab=tests&status=failed

PR #590, opened 2026-03-25
KT-84955: add a supressor for gradle plugin error checker for native target deprecated diagnostic

mMaxy:


gradle.properties
```diff
@@ -20,3 +20,5 @@ org.gradle.kotlin.dsl.allWarningsAsErrors=true
 
 # Temporarily disabled, for details see: #483
 org.gradle.configuration-cache=true
+
+kotlin.internal.suppressGradlePluginErrors=DeprecatedKotlinNativeTargetsDiagnostic
```
fzhinkin, line 24: @mMaxy, should we also set the property (and suppress the deprecation) for all integration test projects (`./integration-testing/examples/`) too?

mMaxy: Yes, probably we should. Added, pushed

fzhinkin (approved):

PR #592, opened 2026-03-30
Fill Implementation-* attribibutes in JAR's manifests

### LINKED ISSUE #568: Fill Implementation-Title/Version/Vendor attributes in JAR's Manifest
Having this information will help with improving DX in IDE.

See https://youtrack.jetbrains.com/issue/KT-83393

fzhinkin:
Closes #568

### DIFF
--- a/buildSrc/src/main/kotlin/JarManifest.kt
+++ b/buildSrc/src/main/kotlin/JarManifest.kt
@@ -1,0 +1,15 @@
+import org.gradle.api.Project
+import org.gradle.jvm.tasks.Jar
+
+internal fun Project.configureImplementationJarManifest(taskName: String) {
+    tasks.matching { it.name == taskName }.configureEach {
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/JarManifest.kt
```diff
@@ -0,0 +1,15 @@
+import org.gradle.api.Project
+import org.gradle.jvm.tasks.Jar
+
+internal fun Project.configureImplementationJarManifest(taskName: String) {
+    tasks.matching { it.name == taskName }.configureEach {
```
shanshin: `matching` is not lazy operation, so it's better to write something like
```
tasks.withType<Jar>.configureEach {
    if(name != taskName) return@configureEach
    ... 
}

```

shanshin (approved):

